PR #87, opened 2024-06-12
adding the profile page

sprinklepancake:
this contains only the profile.vue page that handles the redirection of the user based on their login status


mcfarhat:
Nice work Hassan, let's change the text "Checking login status (^^)" into the standard loader/spinner icon we use across the web solution.
You can find similar usages in the format <i class="fas fa-spin fa-spinner" v-if="loading"></i> 
You might want to also use class "text-brand" or not depending how it looks on the empty page

mcfarhat:
@sprinklepancake one thing that still needs to be done, whenever we have textual content, it needs to be created as a translation text. In your code:

- text: 'You currently do not have any pending rewards',

you need to use a similar approach to the below where you define a new token, and the token text gets defined in the en_US.js  language file
//this.$t('no_pending_rewards'),

sprinklepancake:
done!

PR #91, opened 2024-06-24
adding the swap token functionality

sprinklepancake:


mcfarhat:
hey @sprinklepancake , nice work! few adjustments to be made:
Change the hardcoded values 'actifit.swap', 'AFIT', and partner percentage ... to values configured under nuxt.config and use their mapping in the wallet instead.

mcfarhat (approved):

PR #94, opened 2024-06-28
fixing the breaking modals

sprinklepancake:


mcfarhat (changes requested):
hey @sprinklepancake , you need to keep the modal-dialog class otherwise the modal will not close.
Also it's best if you actually make (or use an existing) component for this modal, instead of coding it into the wallet directly. You could for example use the NotifyModal component and pass it some arguments, or just build a component for the token swap modal.
Please adjust, review, test and submit with changes.

sprinklepancake:
hey @mcfarhat , i did the changes and made a new component file and gave the modal the specified class.

sprinklepancake:
@mcfarhat the latest commit is regarding the single post style issue, i realized that in the activity page it was unchanged so i fixed it.

mcfarhat (approved):

mcfarhat:
refactor modal code as component + fix single post activity view

PR #96, opened 2024-07-04
adding the loader

sprinklepancake:
@mcfarhat added the loader before the iframe shows up

### DIFF
--- a/components/SwapTokenModal.vue
+++ b/components/SwapTokenModal.vue
@@ -1,35 +1,61 @@
-<!-- Swap Tokens Modal -->
 <template>
-    <div class="modal fade" id="swapTokenModal" ref="swapTokenModal" tabindex="-1">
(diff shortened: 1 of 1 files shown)


components/SwapTokenModal.vue
```diff
@@ -1,50 +1,76 @@
...
+          </button>
         </div>
+        <div class="modal-body">
+          <div class="loading-container" v-if="loading">
+            <div class="text-center">
+              <i class="fas fa-spinner fa-spin text-brand"></i>
+            </div>
+          </div>
+          <iframe
+            v-show="!loading"
```
mcfarhat, line 18: is there a reason you're using v-show instead of v-else? i suppose this case is the else of the above (loader/swap)?

mcfarhat (approved):

PR #108, opened 2024-09-21
fixing the cutoff elements in display

sprinklepancake:


sprinklepancake:
this PR contains both changes regarding cutoff elements and the editReportModal.

PR #113, opened 2024-10-17
making the login modal look right on mobile screens

sprinklepancake:


mcfarhat:
temporarily merging those till we find original issue cause

PR #116, opened 2024-11-18
moving away from aws + comments loader fix

sprinklepancake:


mcfarhat (changes requested):
@hassan let's make sure the comment loader is as per requirements on trello (left you a note there)

mcfarhat (approved):

PR #119, opened 2024-11-26
claiming rewards balance fix

sprinklepancake:


mcfarhat:
hey @sprinklepancake , sorry just checking this.
No the change you did does not make sense. you are calling the function which calculates the account value, not the one which updates whether the user has claimable rewards or not.

mcfarhat (approved):

PR #121, opened 2024-12-07
claiming rewards balance fix

sprinklepancake:
added a function that handles wait before fetching user data so i can call it inside the claimrewards function to deal with the blockchain delay.


mcfarhat:
@sprinklepancake you overcomplicated this :)
fetchUserData itself needs to be called the first time after 3 seconds, otherwise the data will not include the correct balance.

sprinklepancake:
hey @mcfarhat , how about this much simpler approach?

mcfarhat (approved):

PR #124, opened 2024-12-30
hover card functionality expanded to more instances on the website

sprinklepancake:
all of them work, but for the report modal on the activity page we are having some issues (mentioned on trello)


mcfarhat (changes requested):
@hassan a nice way to improve this is to include the actual user name, profile display, and link inside the component itself, so you don't need to insert additional details inside the UserHoverCard every time you want it added, and you can pass the details (username as params to the component)

sprinklepancake:
interesting, will look more into it!

sprinklepancake:
@mcfarhat tried something like this, thoughts?

mcfarhat (approved):

PR #128, opened 2025-01-09
fixing the bugs regarding user hover card

sprinklepancake:


pages/yieldfarming.vue
```diff
@@ -83,7 +83,6 @@
 		  <div class="row" v-for="(trx, index) in delegationTransactions" :key="index">
 			<span class="col-6 p-1">
 				<a :href="trx.user" target="_blank" >
```
mcfarhat, line 85: after removing the internal div, is there a useful reason to keep the a href without any content?

components/Comments.vue
```diff
@@ -151,8 +151,9 @@
 		</transition>
 		<div class="comment-reply" v-if="responsePosted">
 			<a :href="this.user.account.name" target="_blank">
+				<UserHoverCard :username="user.name"/>
 			  <div class="comment-user-section" :style="{ paddingLeft: (depth + 1) * indentFactor + 'px' }">
```
mcfarhat, line 155: after removing the card from within, is this div being used for anything now?

mcfarhat (changes requested):
@sprinklepancake  please check notes

pages/yieldfarming.vue
sprinklepancake: not really, i kept it for future reference if we needed it later on. want me to remove it?

mcfarhat: yes please, blank a could cause issues and serves no purpose

sprinklepancake: alright deal! will also remove the one for the other comment you made.

just a couple minutes to properly test it to fully make sure nothing breaks

sprinklepancake:
hey @mcfarhat i removed them and while doing so some bugs came up so i fixed them as well. for some reason the profile pictures for the instances in the yield farming page only show when the class is yield-avatar. so i tried to work around that.

mcfarhat (approved):

PR #134, opened 2025-01-13
making user hover card clickable on mobile

sprinklepancake:


mcfarhat (changes requested):
@sprinklepancake again i can see all those useless a tags, the event should come from within the userhovercard, so you need to actually handle clicks or hover actions from the UserHoverCard component, and get rid of all those a tags

sprinklepancake:
removed them, was testing if their removal affects the functionality and it doesnt, so this should work!

sprinklepancake:
did some resolving to ensure the changes i intended to be there are actually there, this should be working!

sprinklepancake:
saw some issues with the code as well as merging so fixed them again

mcfarhat (approved):

PR #168, opened 2025-06-03
Leo marida

leo-marida:
fixed the issue of accepting invalid input while signing up

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,18 @@
+{
+  "configurations": [
+    {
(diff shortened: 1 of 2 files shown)


mcfarhat (changes requested):
@leo-marida please don't bundle diverse changes into a single PR. PR should be focused on a single task.
Also include a useful title about what this PR covers as you did in the commit title

leo-marida:
Hello doctor. Apologizes for that I didn't know that. I tried actually to
separate them put I couldn't so if I will start doing that staring from the
following task.

PR #169, opened 2025-06-04
Fix signup pattern to prevent invalid usernames(patterns) update

gina2l:
updated the pattern to handle more cases invalid username patterns

### DIFF
--- a/pages/signup.vue
+++ b/pages/signup.vue
@@ -401,6 +401,21 @@ export default {
     async handleUsername(val) {
       this.username_invalid = '';
       this.username_exists = '';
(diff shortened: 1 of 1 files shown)


mcfarhat (changes requested):
much better now @gina2l, though there are cases not covered yet, such as for example uid-11111 (- is causing troubles for your regex)
you do NOT need to create a new PR, just push changes to this same one

mcfarhat (changes requested):
much better now @gina2l, though there are cases not covered yet, such as for example uid-11111 (- is causing troubles for your regex)
you do NOT need to create a new PR, just push changes to this same one

PR #172, opened 2025-06-04
Fixed hamburger icon in navbar in dark mode while using mobile

leo-marida:


components/VoteProposalModal.vue
```diff
@@ -96,6 +96,9 @@ export default {
 </script>
 
 <style scoped>
+body{
```
mcfarhat, line 99: same change as other PR. Should not be there

mcfarhat:
make sure to ONLY include changes relevant to your PR.

mcfarhat (changes requested):

PR #173, opened 2025-06-04
Fixed navbar icons in mobile view where the entire container now and not only the icon lead to the destined page

leo-marida:


mcfarhat (changes requested):
same, includes all changes from other PRs including the bugs mentioned before

PR #174, opened 2025-06-04
Fixed bug in sign up page

leo-marida:
even though they might seem 3 pages but they serve 1 purpose


mcfarhat:
closing for redundancy

PR #177, opened 2025-06-05
fixed sign up page text contrast issue

leo-marida:
This was a bug that i discovered and wasn't a task on trello (details on trello)


mcfarhat (changes requested):
@leo-marida change looks good, you don't need though the extra div, you can set the new class into the .container div

mcfarhat:
closing for redundancy

PR #180, opened 2025-06-05
Username invalid pattern

leo-marida:
same task that me and ghina were working on which is also on trello


mcfarhat (commented):
Overall looks fine, couple of recommendation: 
- avoid hardcoding text into code. Text should always be placed under language files (en_us.js)
- try to make the check for username validity in a single hit (checking for both uid and sequence of numbers)
Closing PR since Ghina's PR already merged

mcfarhat:
Overall looks fine, couple of recommendation: 
- avoid hardcoding text into code. Text should always be placed under language files (en_us.js)
- try to make the check for username validity in a single hit (checking for both uid and sequence of numbers)
Closing PR since Ghina's PR already merged

PR #181, opened 2025-06-05
Signupfinalchange

gina2l:


mcfarhat (changes requested):
This still includes the steemconnect patch. the en_ur and signup files are fine, but you dont need ot have vue-steemconnect+0.3.1.patch committed

PR #184, opened 2025-06-06
user menu now fixed for mobile view and pc view and search bar size fixed and count fixed as well

leo-marida:


mcfarhat:
> > what is this PR for? it's not referenced in trello
> 
> it's for the issue of the notification bell that wasn't on a straight line in the mobile view and it is on trello. I fixed this issue but you highlighted that the icons are quite small as well as the search bar was also small so i fixed those in this pr. You mentioned something as well related to the number of notification in both pc and mobile views so i tried to fix it as well. This is the pr: #182

So this is a replacement PR? if so, please mention it, close the old PR, and update trello.

mcfarhat:
looks good

PR #187, opened 2025-06-10
Sidebar post incomplete 

gina2l:
the links in the sidebar to the user info are not working


mcfarhat:
@gina2l few notes:

1- the tops are till not aligned, left sidebar upper border is below the upper border of the post
current view:
image
expectation:
image

2- when you scroll up, the sidebar hides behind the menu, this should not be the case
image

3- you used a different user rank icon than the one used through the app. Let's stick to the current one
image

4- i would also expand the whole screen minimizing white margins on the side (areas in red on the sides, make them smaller by pushing sidebar more to the left, and expanding the area of the post content, maybe using rate 1/3 for sidebar and 2/3 for post)
image

5- in mobile view, the post content should show before the sidebar not after it
image
image

PR #191, opened 2025-06-11
fixed user menu icons.

leo-marida:
fixed all the changes that you have requested


mcfarhat (changes requested):
Changes look good. Please restore icon sizes in desktop to original ones.

PR #209, opened 2025-06-23
issue fixed.

leo-marida:
i noticed that the colors of the steem are perfect together and thus there was no need for the hovering and thus i removed the css effect when hovering this component


mcfarhat (approved):

mcfarhat (approved):
color contrast issue in main box on landing page

PR #291, opened 2026-07-01
I removed the white thick white margin from the notification.

ShahdQaddoura:


mcfarhat (changes requested):
Thanks @ShahdQaddoura! The dark-mode modal styling is a nice improvement. A few things to address before this can merge — assigning back to you:

## 🔴 Blocker — `package-lock.json` changes must be reverted
This PR removes `undici@6.26.0` and `buffer@6.0.3` entries from `package-lock.json`. That looks like an unintended side effect of running `npm install` on a different npm version, not a deliberate change.

Production deploys with `npm ci` under **Node 16.20.2 / npm 8.19.4**, and `package.json` / `package-lock.json` must stay perfectly in sync or the build fails at install with `EUSAGE`. PR #284 was specifically made to *sync* these undici/ws lockfile versions for `npm ci` — this change risks re-breaking that.

**Fix:** restore `package-lock.json` to develop's version so this PR only contains the two `.vue` files:
```
git checkout develop -- package-lock.json
git commit -m "revert unintended package-lock changes"
```

## 🟠 `NotifyModal.vue` styling isn't scoped to dark mode
The new rules are unconditional:
```css
#notifyModal { background-color: black !important; }
#notifyModal .modal-content { background-color: black !important; color: white !important; }
```
This forces a black modal with white text in **all** modes — so in light mode the notification becomes white-on-black, which looks broken.

Your `VoteProposalModal.vue` change does this correctly by scoping everything under `.dark-mode`. Please do the same here — prefix these rules with `.dark-mode` so light mode is unaffected:
```css
.dark-mode #notifyModal { ... }
.dark-mode #notifyModal .modal-content { ... }
```

## 🟢 `VoteProposalModal.vue` — looks good
Properly scoped under `.dark-mode`, sensible overrides for background / text / header / close button. No changes needed here.

Once the lockfile is reverted and NotifyModal is dark-mode-scoped, this is good to merge into develop. 🙏

PR #306, opened 2026-07-13
Fixed suggested friends rendering, loading and styling

ShahdQaddoura:


mcfarhat (changes requested):
Thanks @ShahdQaddoura — the Suggested Friends fix itself looks good: the panel renders properly now with avatars, activity/AFIT/mutual-friend counts and per-stat loading spinners. 👍

**One thing to fix before this can merge: this PR bundles two unrelated changes.**

Looking at the 5 changed files, there are actually two independent pieces of work here:

1. **The advertised fix — Suggested Friends** (`components/FriendshipModal.vue`). This is self-contained and uses its own endpoints (`userFriends`, `userFriendRequests`, `addFriend`, `performTrx`).

2. **A separate auth→proxy migration** (`api/proxy.js`, `components/LoginModal.vue`, `pages/login.vue`, `pages/_username/wallet.vue`) — moving the login / keychain / captcha calls behind the server-side proxy (new `/loginAuth`, `/loginKeychain`, `/loginKeychainVerify`, `/verifyLoginCaptcha`, `/getAccountData` handlers).

These share **no code path** — the login work isn't needed for Suggested Friends, and FriendshipModal doesn't call any of the new proxied endpoints. So this is two features in one PR.

**Please split them:**
- Keep this PR (#306) scoped to **just the Suggested Friends fix** (`FriendshipModal.vue`).
- Move the **auth→proxy migration** to its **own Trello card** and its **own branch off an up-to-date `develop`**, as a separate PR.

Rule of thumb going forward: **one PR = one concern.** Two unrelated new pieces of functionality shouldn't ride in the same PR/branch — it makes review harder and couples unrelated changes (e.g. an auth issue could block a simple UI fix, or vice-versa).

**For the separate auth-proxy PR** (the migration is a good idea — it aligns with our proxy security model — it just needs two things before it ships):
- **Add rate limiting** to the new login endpoints via `isRateLimited(...)`, like every other endpoint in `api/proxy.js` does. These are auth endpoints, so throttling matters most here.
- **Cap the request body size** in `readRequestBody` — right now `body += chunk` is unbounded, which is a memory-exhaustion risk. Abort once it passes a sane threshold.

Once #306 contains only the Suggested Friends change, it's an easy approve. 🙏

PR #312, opened 2026-07-17
Fix error sending messages

reemnii:


mcfarhat (changes requested):
Thanks @reemnii — solid work on the send-message fix. The Keychain/PeakVault path is much better now: the `handleWithKeychain` override that preserves the recipient username for `requestEncodeMessage` is the right root-cause fix, and the readable-error handling (no more `[object Object]`) + `beforeDestroy` cleanup + script-load dedup are all good improvements.

Before this can merge, there's one likely **regression** plus a few smaller items:

## 🔴 Posting-key login users can no longer chat (regression)
Actifit supports **posting-key login** (standard login), and those users could previously **send chat messages** via the Sting widget. The old component deferred to the widget's own login: `requireLogin: false` + `setUser(username)` (no method) with StWidget's built-in keychain/login passthrough left enabled — so the widget handled auth itself, including posting-key sign-in.

The new code changes that:
```js
requireLogin: !username || (!hasKeychain && !hasPeakVault && !localPostingKey),  // true for non-wallet users
...
} else if (username) {                       // no keychain / peakvault / (unset) postingKey
  this.configureUnavailableWalletMessage()    // actively BLOCKS every wallet request
  this.widgetError = 'Private messages require Hive Keychain or PeakVault.'
}
```
So posting-key (and HiveAuth) users now get "Private messages require Hive Keychain or PeakVault" and **can't send** — where they could before. Please restore support for posting-key users.

## 🔴 The `chatPostingKey` path is a dead stub
The branch meant to cover exactly those users reads `this.$store.state.chatPostingKey`, but **nothing in the app ever sets `chatPostingKey`** (and we don't persist a posting key anywhere), so `setPostingKey(...)` never runs. If posting-key chat is meant to work, this needs to be wired to the logged-in user's posting-key session; otherwise it's dead code. This is almost certainly why the regression above happens.

## 🟡 Hardcoded English error strings
"Chat is temporarily unavailable…", "Private messages require Hive Keychain or PeakVault…" should use `$t()` — the app ships 14 locales.

## 🟡 Verbose console logging left in
`console.info('[StingChat] …')` for wallet availability/requests/responses. It doesn't leak keys or message content (good), but please trim or gate it for production.

## Heads-up — #311 is now merged, please rebase
The message-bar cropping fix (**#311**) has been reviewed, tested, and **merged into develop**. Since #312 also edits `StingChat.vue`'s `<style>` block — it carries the same cropping CSS, just under the renamed `.sting-widget-container` selector — you'll get a small conflict when you bring develop into this branch. Resolution is straightforward: **keep your `.sting-widget-container` version** of those rules (your new template renames the container, so develop's `#widgetContainer` selectors no longer apply) and drop the duplicate `#widgetContainer` block. Please **rebase this branch on the latest develop** as part of the rework.

Once posting-key users can send again (and the `chatPostingKey` path is either wired up or removed), and this is rebased on develop, it's good to go. 🙏

mcfarhat (changes requested):
Thanks @reemnii — this rework is excellent and addresses everything from the first round: posting-key chat is restored, `chatPostingKey` is properly wired (login capture → memory store → prompt fallback), the key is memory-only + cleared on logout + validated against the account's `key_auths`, i18n across all 14 locales, logging trimmed, and you added unit tests (all passing) and rebased cleanly on develop. Really solid.

However, hands-on testing surfaced **two blockers** — and they share one root cause, so one fix resolves both.

## Root cause: the widget is created once and never reconfigured/torn down when the active user changes
`initWidget()` runs once in `mounted()`, and there is **no `watch` on `user`**. Combined with the `chatPostingKey` watcher only reacting to a *new* key:
```js
watch: { chatPostingKey (postingKey) {
  if (!postingKey || !username || !this.widget) return   // early-returns when the key is CLEARED
  this.configurePostingKey(username, postingKey)
}}
```
…nothing tears down the widget's signer state on logout/switch.

### 🔴 1. Security: old user's key stays active after switching accounts
On logout/switch the store nulls `chatPostingKey`, but the watcher early-returns, so the widget's `handleWithHivejs` **closure still holds the previous user's posting key** and keeps signing. Result: after switching users, chat can still post/sign as the **previous** account. This is the critical one.

### 🔴 2. Switching to Keychain shows no confirm popup and hangs the chat
Because the widget was first configured for a posting-key user, it set `enableKeychainPassthrough = false` and installed the local posting-key signer. After switching to a **Keychain** account the widget is still in that state — so `handleWithKeychain` is never invoked (**no Keychain popup**), and Sting's request never receives a response, so the **chat display hangs** on a promise that never resolves.

## Fix direction
Reset the widget whenever the active user changes **and** when `chatPostingKey` goes null:
- Add a `watch` on the account name (`user.account.name`). On change — and in the `chatPostingKey` watcher's null branch — tear down the current signer: clear the `handleWithHivejs` / `handleWithKeychain` overrides, restore `enableKeychainPassthrough` / `enablePeakVaultPassthrough` to their defaults, then re-run `initWidget()` for the new session (simplest robust approach: `this.widget.cleanup()` + recreate). A stale signer closure must never survive a logout/switch.
- Please also verify the **Keychain path works on a fresh Keychain login** (not only after a switch), in case there's a second issue in the passthrough itself — the "no popup + hang" symptom is worth confirming independently.

Everything else looks good — just these two (one fix) before it can merge. Returning your Trello card to Doing. Thanks! 🙏

mcfarhat (commented):
Re-reviewed the "Fix chat session handling" update, @reemnii — this correctly resolves both blockers, and I like how you did it:

- **Stale-signer security bug fixed:** `teardownWidget()` now nulls `handleWithHivejs`/`handleWithKeychain` (dropping the closures that held the previous user's posting key) before cleanup, and a `watch` on `activeUsername` + the `chatPostingKey` null-branch both call `restartWidgetSession()` — so the old key can no longer survive a logout/account switch.
- **Keychain-hang fixed:** restoring `enableKeychainPassthrough`/`enablePeakVaultPassthrough` to `true` on teardown means a subsequent Keychain login gets its popup and Sting's request resolves.
- The `widgetSessionVersion` / next-tick recreate guards against a stale async callback from a torn-down session — nice touch. And thanks for the unit tests.

Verified by reading the diff; the fix matches the recommended direction. Merge-ready from my side — holding the merge for Dr. Mohammad's sign-off. (One thing to confirm on your end per the last note: a *fresh* Keychain login shows the popup, not just after a switch.) Great work.

mcfarhat (approved):
Approving after the full re-review — cloned, installed, ran the test suite green, and a security pass came back clean. Details in the review thread.

PR #318, opened 2026-07-24
Added option to edit body measurements

reemnii:


mcfarhat (commented):
Really nice work here, @reemnii — and thank you for adopting the clean branch flow. This is exactly right:

- **Proper feature branch** (`Edit-body-measurements → develop`), one focused commit — a big improvement over the previous branch tangle.
- **Well-factored logic.** The measurement helpers in `utils/measurements.js` are pure and defensive — unit-alias normalization (centimeters/kgs/pounds → cm/kg/lb) with a sane default fallback, cm↔in and kg↔lb conversion with non-finite/same-unit guards and 2-decimal rounding, and `normalizeMeasurements` validating each value before use. Bodyfat correctly treated as unitless (%). 
- **Genuine unit tests** — conversions both directions, alias + default handling, merge-in-date-order, and metadata preservation. Great to see tests land with the feature.
- **Solid UI.** Edit gated on `canEditMeasurements` (owner-only), numeric inputs with `min="0"`/`step="any"`, a double-submit guard (`savingMeasurements`), inline error display, and a cancel path. Full i18n across all 14 locales too.

One tiny, non-blocking note: this repo has no CI to auto-run the spec, so please confirm `npm test` passes locally before it goes in (I expect it does). Optional: in `convertMeasurementValue`, the cm↔in branch isn't gated on the field, so a mismatched weight+cm would misconvert — it can't happen through the UI, but a `key`/unit sanity check would make the helper bulletproof.

Overall this is merge-ready from my side — holding the actual merge for Dr. Mohammad's sign-off. Great comeback on this one. 👏

mcfarhat (changes requested):
Following up on my review, @reemnii — I did a deeper pass on the persistence layer, and it's implemented well (on-chain via `account_update2` into `posting_json_metadata.profile.actifit_measurements`, other profile fields preserved, no keys in code — nicely done). Two things to fix before this merges, plus two quick confirmations:

## 🔴 1. The `account_update2` can wipe the account's `json_metadata`
In `saveMeasurements` you broadcast:
```js
json_metadata: this.userinfo.json_metadata || '',
posting_json_metadata: JSON.stringify(postingMetadata),
```
`account_update2` overwrites whatever fields you pass. You deep-merge `posting_json_metadata` correctly, but `json_metadata` is passed raw with an `|| ''` fallback — so for any account where `this.userinfo.json_metadata` isn't the live on-chain value (or is empty while the account actually has `json_metadata` set), this **overwrites their `json_metadata` with an empty string** and loses whatever was there.
**Fix:** never default it to `''`. Either omit `json_metadata` from the op entirely so only `posting_json_metadata` is updated, or pass through the freshly-fetched on-chain `json_metadata` unchanged. Please check the other broadcast block too (the `pst` one) — it has the same `json_metadata: ... || ''` pattern.

## 🔴 2. "Latest" display shouldn't let a stale profile edit override a newer report
Right now:
```js
const latest = { ...reportLatest, ...(profileMeasurements || {}) };
```
means the profile edit **always** wins, regardless of dates. So if a user edits their profile and then later logs a **newer** activity report, they'll still see the older manually-edited values as "latest" (and `lastUpdated` shows the profile's date, not the newer report's). For a fitness app the newer measurement should win.
**Fix:** make the "latest" selection **date-aware** — the profile value should win only when `profile.updated_at` is newer than the latest report's date; otherwise the report wins. You already sort correctly in `mergeMeasurementSources` (date desc), so one clean option is to derive the displayed "latest" from that sorted list rather than the unconditional spread — just keep the gap-filling behavior (fall back to the report for any field the winning source doesn't have) and make `lastUpdated` reflect whichever source actually won.

## 🟡 Two quick confirmations (no change if already true)
- Confirm `$processTrxFunc('account_update2', …)` signs with the **posting** key, not active — a profile edit shouldn't require the active authority.
- Confirm `getProfileMetadata()` safely handles a malformed/empty `posting_json_metadata` (try/catch → `{}`) so a bad existing metadata string can't throw mid-save.

The feature itself is solid and the tests are great — these are about not clobbering existing on-chain data and getting "latest" right. Ping me when they're in and I'll re-review. Moving the card back to Doing meanwhile. Thanks!

mcfarhat (commented):
Re-reviewed after your update, @reemnii — both items are fixed correctly, nicely done:

- **json_metadata no longer clobbered:** `getLiveAccountJsonMetadata()` fetches the account's current on-chain json_metadata and passes it through in both broadcasts. Great that you added a test for the exact edge case (empty local `userinfo.json_metadata` still preserves the live value) — that's the failure I was worried about.
- **"Latest" is now date-aware:** `selectLatestMeasurements` takes each field from the newest source and gap-fills from older ones, and `lastUpdated` reflects the winning source. Exactly right — a newer activity report is no longer hidden behind an older profile edit.

Verified the logic by reading the diff; your added tests cover both cases. This is merge-ready from my side — holding the merge for Dr. Mohammad's sign-off. Solid turnaround.

mcfarhat (approved):
Approving after the full re-review — cloned, installed, ran the test suite green, and a security pass came back clean. Details in the review thread.

PR #338, opened 2026-07-27
Fix: refresh VP and RC on user switch

reemnii:


mcfarhat (approved):
Approving — reviewed the full `SteemStats.vue`, not just the diff. This is a correct, well-tested fix: VP is assigned synchronously before the `getRCMana` await, and the monotonic `rcRequestId` guard (plus the `username` check) correctly discards out-of-order/previous-user RC responses. Solid coverage of user-switch, logout, overlap, chain-switch, and unmount, and `beforeDestroy` fixes the real `setInterval` leak. Nice work, @reemnii.

One process note for next time: please add a **description to the PR body** (what the bug was + how you fixed it). This one shipped with an empty body — the summary's going into the merge commit this time, but going forward a short description on every PR makes review and history much easier. Merging now.

PR #340, opened 2026-07-27
Edit body measurements

reemnii:
Posting-key saves failed because `account_update2` required active authority, while Keychain rejection caused a null transaction result crash and locked the form. Fixed the metadata payload to use posting authority and added safe error, cancellation, timeout, and transaction-response handling.


mcfarhat (approved):
Approving — strong work, @reemnii. The account_update2 with json_metadata:'' (no-change sentinel) to keep the profile update on POSTING authority is exactly right, plus the robust error extractor, the save-timeout wrapper, and solid test coverage. One minor follow-up (tracked on the card, non-blocking): the Keychain 'user_cancel' response surfaces untranslated when a user cancels the edit — worth mapping to a localized message (or silently ignoring, since a cancel isn't an error).

PR #342, opened 2026-07-27
fix(i18n): translate profile and measurement labels across all locales

reemnii:
Fixes raw i18n keys on the user profile page by adding missing profile, measurement, and activity translations across all 14 locales, with regression coverage.


mcfarhat (approved):
Approving — clean, additive i18n fix: adds the missing profile/measurement/activity keys across all 14 locales with a parity regression test, no removals. Thanks @reemnii. (This also provides the strings that PR #340 references.)

PR #343, opened 2026-07-27
added 'logout' and 'Switch User' button on the profile page.

ShahdQaddoura:


mcfarhat (approved):
Approving and merging, @ShahdQaddoura — clean, self-contained addition (Switch User + Logout on the own-profile view, guarded by isOwnAccount()). Two non-blocking follow-ups are tracked on the Trello card: (1) the buttons need a dark-mode variant — they currently render identical in dark since there's no html.dark-mode rule for them; (2) a small test for proceedLogout would be good. Nice work.

PR #364, opened 2026-08-02
Feature/revamp market screen

reemnii:
Revamps the Actifit Market with a responsive two-panel layout, improved product and pricing accuracy, search and advanced filtering, clearer ownership/status information, and saved products.


mcfarhat (changes requested):
Thanks @reemnii — the market revamp is a big step up, but it needs another pass before it can merge. Please address the following:

## 🔴 Failing tests (must be green)
Running the PR's own market suites: **4 tests fail** in `test/unit/pages/market-owned-gadgets.spec.js`.
- The tests call `MarketPage.methods.ownedGadgetCount` and `MarketPage.methods.activeGadgetCount`, but **neither method exists** in `pages/market.vue` — so the calls throw / return wrong values.
- `computed.groupedProducts` does not produce the expected grouping: the test expects a leading **saved-products** group and **owned-first** ordering, but the code returns type-based (`ingame`) groups instead.

A PR should never ship with its own tests red — either implement the missing methods/grouping, or update the tests to match the intended final behavior. (The other 6 market suites pass: 85/89 total.)

## 🟠 Testing feedback (from a local QA pass)
1. **Empty categories should not render.** `no_products_found` is showing under the *Consultation* category. If no products match a category, hide the category entirely rather than showing an empty-state under it.
2. **Cart/checkout needs a clear global entry point.** We support a cart now, but checkout access only exists inside individual products. Add a visible cart/checkout affordance outside products — e.g. next to the sort icon.
3. **"Available" indicator is misleading.** The availability circle shows on *every* product, including ones the user cannot buy. Rework it so it only reflects true buyability (don't imply a product is available/purchasable when it isn't).
4. **Owned/bought products aren't categorized.** After buying a product for a user, it does not appear grouped under that user's owned/bought items (as we discussed previously). Please add an easily accessible separation of products by state: **bought / activated / available / unavailable (requirements not met)**.

Happy to re-review once the tests pass and these are addressed. 🙏

mcfarhat (changes requested):
## Round 2 — retested with the QA-feedback commit (085d576)

Nice progress @reemnii — verified fixed: **all 94 market tests pass**, empty categories are hidden, the global cart button + modal work, and status dots are now state-driven. A few more issues from a fresh QA pass (desktop + mobile) before this can merge:

### 🔴 i18n
1. **Empty-state shows the raw key `no_products_found`.** It's referenced at `pages/market.vue:257` via `$t('no_products_found')` but **not defined in any of the 14 `lang/*.js` files**, so users see the literal string. Add the key to all 14 locales. (This is the same thing that showed under Consultation originally — the category-hiding is fixed, but the untranslated key still shows in the true empty state.) Please also add a regression test like `market-labels.spec.js`.

### 🔴 Left-side menu (sidebar) consistency
2. **Gadget Level (L1/L2/L3) color-coding** was fixed in the product detail view but **not in the left-side menu rows** — make them consistent.
3. **Christmas / special-event gadgets should not show inline** in the normal list. They should only appear **when activated** (they're currently inactive, so should be hidden), and should live in a **separate "Event" section**.

### 🔴 Mobile
4. **Filters tab:** opening it highlights **both "ALL" and "Ebook" in red** at once — only the genuinely active filter should be highlighted.
5. **Can't scroll the product list** — the view feels stuck/trapped, you can barely move past the current items (looks like an inner/nested scroll container capturing the scroll). The list should scroll naturally.
6. **Single-gadget layout is scrambled/misaligned** — gadget type is on one row, Save on another, and the image and text aren't aligned. Needs a cleaner, aligned layout.

### 🟡 Layout polish (desktop + mobile)
7. The DHF "Hive Dev Proposal" popup overlaps the product area (top-right) and the cookie-consent banner overlaps a mid-list product row — check z-index/spacing so the global overlays don't collide with the new market layout.

### Pre-merge reminders
- **Merge `develop` into the branch** — it's behind, and this pulls in the 1.12.1 login fix (the branch currently carries the old pre-fix login code).
- Confirm full `npm test` green. Note: the 2 failing `StingChat.spec.js` tests are **pre-existing on develop (from #358), not yours** — I'm tracking those separately, don't worry about them here.

Ping me when it's ready and I'll re-test. 🙏

mcfarhat (approved):
✅ Approved — round-3 addresses every item from the previous review, verified against the code + live UI:

- **`no_products_found`** now defined in all 14 lang files (no more raw key).
- **Sidebar gadget-level coloring** now uses the shared `getGadgetLevelClass` → consistent with the product detail (rows carry `gadget-level-1/2/3`).
- **Event/Christmas gadgets**: moved to a dedicated "Event" group, and `filterMarketProducts` correctly hides any `specialevent` product unless the user has an **active** unit — so inactive event gadgets no longer show at all. 👍
- **Mobile filters**: only the truly-active option highlights now (Type + Currency sections each show just "All" active by default).
- **Mobile scroll**: list scrolls with the page naturally (sidebar no longer traps scroll).
- **Mobile single-gadget layout**: type badges + Save aligned on one row, image/text aligned.

Tests **126 pass**, lint clean, and `develop` merged in (carries the 1.12.1 login fix).

🟡 Non-blocking follow-up: the DHF "Hive Dev Proposal" popup + cookie banner still overlap the market content — but that's a *global* overlay affecting other pages too, so better handled as its own ticket. Nice work! 🎉

mcfarhat (changes requested):
Reopening — testing at true phone widths surfaced a **horizontal-overflow bug** on mobile that my earlier pass missed (I only checked vertical scroll). Apologies for the churn.

### 🔴 Mobile horizontal overflow (≤ ~360px)
At a 320px viewport the market list overflows sideways — the AFIT price column gets clipped on the right and the page scrolls horizontally.

**Repro:** DevTools responsive @ 320px → `document.scrollWidth = 373` vs `clientWidth = 320`. The `.market-sidebar` / `.sidebar-scroll` / `.sidebar-row` render **356px wide** inside the 320px viewport.

**Cause:** on mobile the layout is `.market-layout { grid-template-columns: 1fr }`, but the grid item `.market-sidebar` has no `min-width: 0`, so the track behaves as `minmax(auto, 1fr)` and refuses to shrink below the row content's min-width (the price column forces it). It only bites ≤ ~360px; ≥ 375px is fine, which is why it slipped through.

**Suggested fix** (in the `@media (max-width: 900px)` block):
```css
.market-layout { grid-template-columns: minmax(0, 1fr); }
.market-sidebar { min-width: 0; }
```
plus let the price column (`.sidebar-row-side`) shrink/wrap instead of forcing width. Please verify at 320px that `document.scrollWidth === clientWidth` (no sideways scroll) and the price column isn't clipped.

Everything else from round-3 is verified good — this is the only blocker. Also worth a quick pass on the other narrow widths (320/360) while you're in there. 🙏

mcfarhat (changes requested):
Re-tested round-4. ✅ The mobile horizontal-overflow fix is confirmed (no overflow at 320 / 360 / 390px), and the round-3 items still hold. Three new issues from this testing pass before it can merge:

### 🔴 1. Garbled Sort dropdown (character encoding)
`pages/market.vue` lines 74–83 now literally contain `â–²` / `â–¼` — the ▲/▼ got double-encoded (the "fix market character encoding" commit corrupted them). They render as "Priceâ–²", "Boughtâ–¼", etc.
**Fix:** replace the literal arrows with HTML entities — `&#9650;` (▲) and `&#9660;` (▼). Those render correctly and won't re-corrupt regardless of editor encoding. (Corruption is confined to these 10 lines; lang files are clean.)

### 🔴 2. Login username autofills the market search
The search input (`market.vue:63`) has no `autocomplete` attribute, so the browser autofills it with the saved username when you use the login field (which is `name="username"`) — the list then shows 0 products.
**Fix:** add `autocomplete="off"` and a distinct `name` (e.g. `name="marketSearch"`) to that input.

### 🔴 3. Rename the "Unavailable" group → "Locked"
Those items aren't unavailable — they're purchasable, the user just doesn't meet the requirements (rank / AFIT / prerequisite). "Unavailable" is misleading.
**Fix:** relabel that group to **"Locked"** (a 🔒 icon would reinforce it). Update the `market_state_unavailable` label value to "Locked" across all 14 lang files (or add a `market_state_locked` key). Keep the semantics the same — only the display text changes.

Everything else is verified good — these three are the remaining blockers. 🙏

mcfarhat (approved):
✅ Re-review passed — all three reported issues fixed (sort arrows via HTML entities, search autofill via distinct name + autocomplete, Unavailable→Locked translated in all 14 langs), the earlier double-encoding is gone (0 mojibake), the saved-products blank-panel concern is resolved, mobile overflow fixed, tests 62/62 + lint clean. Merging. Two minor follow-ups (lock icon + collapsible sections) tracked on the card — not blockers.

PR #376, opened 2026-08-04
Remove Consultants from the Navbar

reemnii:
Removes the Consultants item from the navbar.

### DIFF
--- a/components/NavbarBrand.vue
+++ b/components/NavbarBrand.vue
@@ -75,14 +75,6 @@ href="#news"
             </a>
           </li>
 
-          <li class="nav-item btn btn-brand m-1 nav-item-border" :title="$t('professionals_title')">
-            <a class="nav-link text-white d-flex align-items-center justify-content-center w-100 h-100"
-               href="/consultants"
-               @click.prevent="$router.push('/consultants'); closeMenu()">
-              <i class="fas fa-user-md text-white"></i><span>{{ $t('professionals_title') }}</span>
-            </a>
-          </li>
-
           <li class="nav-item btn btn-brand m-1 nav-item-border" :title="$t('Communities')">
             <a class="nav-link text-white d-flex align-items-center justify-content-center w-100 h-100"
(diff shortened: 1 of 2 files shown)


mcfarhat (changes requested):
Thanks for the quick turnaround @reemnii — the removal itself is clean: the Consultants `<li>` is gone, and New Blog + New Video are correctly kept. Just one thing to fix before this can merge:

## 🔴 This breaks a unit test
`test/unit/seo/internal-navigation.spec.js` still asserts the navbar links to `/consultants`, so removing the item makes that test fail:

```
Tests: 1 failed, 5 passed
● internal navigation › links to /consultants from the primary navigation
  expect(primaryNavigation).toContain(`href="/consultants"`)
```

That test was added by the same PR (#356) that introduced the nav item. Please **remove the `'/consultants'` entry (line 12) from the `it.each([...])` list** in `test/unit/seo/internal-navigation.spec.js` so the suite goes green again. `/blog/new` and `/videos/new` should stay in the list.

## ✅ Otherwise good
- `/consultants` page + the service-product link in `Product.vue` are intentionally untouched — the page stays reachable, only the navbar entry is gone. 👍
- `professionals_title` i18n key is still used by `pages/consultants.vue`, so leaving it is correct.

## 🟡 Nit (non-blocking)
Title/branch spell it "Concultants" — worth fixing the title for tidiness, but not required to merge.

Run `npx jest test/unit/seo/internal-navigation.spec.js` locally to confirm green, then I'll re-review. 🙏

mcfarhat (approved):
✅ Re-reviewed — this is good to merge.

- Consultants `<li>` cleanly removed; New Blog + New Video kept.
- The `internal-navigation.spec.js` fix is correct (dropped `/consultants`, kept `/blog/new` + `/videos/new`), and the suite now passes (5/5).

Thanks for the quick turnaround! Note: to actually remove it from production it still needs a `develop → master` release after merge.

PR #377, opened 2026-08-05
revamp wallet screen to a more user-friendly better UX/UI screen

ShahdQaddoura:


mcfarhat (commented):
Reviewed the wallet revamp — solid work overall. Lint is clean, and this is a **presentation-only** change: no data-fetching or store logic was added or removed, and no wallet actions (transfer/stake/delegate/claim) were dropped — the removed lines are just restructured layout containers. 👍

### ✅ Good
- **`nuxt.config.js`**: adding `|| 'https://api.hive-engine.com/rpc/'` as a fallback for `hiveEngineRpc` is a nice hardening — the wallet reads Hive-Engine balances, so this avoids a broken state when `HIVE_ENG_RPC_NODE` is unset (e.g. local dev).
- Dark-mode readability rule for the holders tables (`#8fe3b0` on first/last cells) is a sensible contrast fix.

### 🟡 Please address — CSS scope leak
`components/GenericTopHoldersModal.vue` changes `<style scoped>` → `<style>` (un-scoped) to make the `.dark-mode .holders-table` selector work. Side effect: the block's **`.token-logo { width:40px; height:40px }`** and the **`@media (max-width:600px) .table td/.table th`** padding rules are now **global**. `.token-logo` is used by ~10 other components (CartModal, Footer, Product, PendingRewardsModal, …) and `.table` is Bootstrap-wide, so this can bleed onto other UI wherever this modal is loaded.

**Suggested fix:** keep `<style scoped>` for the block, and move **only** the intentional `.dark-mode .holders-table …` rule into a separate non-scoped `<style>` block (it's already namespaced by `.holders-table`, so it stays safe). That keeps `.token-logo`/`.table` scoped. (`TopHoldersModal.vue` was already un-scoped, so that one's fine.)

### 🔎 Please functionally smoke-test before merge
The wallet template was heavily restructured (+590 lines), so please verify the interactive paths still work end-to-end (logged in via Keychain/HiveAuth): token balances render, **transfer / stake / unstake / delegate / claim rewards**, savings, transaction history, and the top-holders modals — plus mobile layout.

### Pre-merge
- Merge `develop` in (pulls the 1.12.1 login fix; this branch may carry pre-fix login code).

Nice, clean PR otherwise — just the scoping fix + a functional pass and it's good to go. 🙏

mcfarhat (changes requested):
Formally marking changes-requested to reflect status (see my detailed review above): re-scope GenericTopHoldersModal.vue's <style> so `.token-logo`/`.table` don't leak globally, and functionally smoke-test the restructured wallet. Merge develop in for the 1.12.1 login fix.

mcfarhat (changes requested):
Requesting changes — please address the detailed feedback and the 11 annotated screenshots left on the Trello card: https://trello.com/c/fH8FGglj

The card/tile direction is improving, but several items still need work (full context + screenshots on Trello). Summary:

- **Action boxes** (Delegations, Claimable Rewards, etc.) — layout looks bad; needs cleanup.
- **Transaction history** — the 4-boxes-per-row layout is confusing for history. Switch back to **one box per row**, and add a second display for the **Hive-chain financial transactions** so AFIT and Hive-chain histories sit side by side.
- **Filter funnel** no longer turns green on click (unclear if it's working); the **Save settings** button should become a **save icon** and is barely visible.
- **Top bar** — show the token icons next to each balance (AFIT, HP), and keep it consistent (include the "HP" unit on the HP balance like you do "AFIT" on the AFIT balance).
- **Power-down** (expanded) looks bad — improve.
- 🔴 **Top HBD holders / Top HIVE holders** — is this new? The **modal is erroring out** for both. Please fix (functional, not cosmetic).
- **Action-button boxes** are still outdated and should appear **right below the action**, not after scrolling all the way down.
- **AFIT is off-chain** — the main AFIT balance must not be grouped in the same on-chain category.
- **Header alignment** — "Staked" (and sometimes all headers) is misaligned.

This is in alignment with the feedback on the Trello card — please work from there (annotated screenshots for each point). 🙏

mcfarhat (changes requested):
Reviewed at `46b9e7d1`. Two independent passes, both standing the component up and driving it rather than reading the diff, plus live calls to the syncad and Hive APIs. I verified the load-bearing claims myself.

**Both things I asked for are genuinely fixed**, and the modal fix is mutation-proven: reverting to `response.data.map(...)` reproduces `TypeError: response.data.map is not a function`, and with your change the modal renders 100 rows. The values are right too — cross-checked against `condenser_api.get_accounts`, `cold.dunamu` renders `178,799,753.946 HIVE` which matches its on-chain balance exactly. The `|| []` guard was a nice addition: a future shape change now degrades to the empty state instead of throwing again.

And **`off_chain` is in all 14 locales with real translations** — Arabic خارج السلسلة, Japanese オフチェーン, Chinese 链下 — not copy-pasted English. The AFIT grouping is structurally sound: I walked the compiled template AST, the headings render with their groups, and `sortTokenData` only touches the H-E list so sorting can't reorder rows across headings.

**Two things worth saying before the list.** You brought the branch current with `develop` — `rev-list --left-right --count` is `0 5`, zero behind — so the "merge develop in" ask from yesterday is done. And the `BLOCKED` status is **not yours**: `develop` requires one approval with `require_last_push_approval=true` and `dismiss_stale_reviews=false`, so my 08-08 review is still standing against an old commit. Nothing mechanical for you to do there; it clears when I approve the current head.

---

## Blocking

### 1. The new quick-action pills render on wallets the visitor doesn't own

`wallet.vue:42` — the new `<nav class="wallet-quick-actions">` has no `v-if="!nonAuthUser"`. Every other action block in the file does: lines 75, 164, 250, 361, 396, 405, 551, 607, 1055, 1124, 1139, 1199. The pattern is consistent throughout, so this reads as one block missed rather than a misunderstanding.

Executed with `user: null` on route username `someoneelse`: `nonAuthUser === true` and all five pills render, including `TRANSFER_FUNDS_ACTION_TEXT`. Clicking it sets `fundActivityMode = 1` and renders the full transfer form with a **Send** button and the amount prefilled from the *viewed* user's balance.

To be precise about severity, because it would be easy to overstate: this is a **UI exposure and a mis-send hazard**, not a way to move someone else's funds — any actual transfer still requires the viewer's own signature. But offering a Send button, prefilled from a balance that isn't yours, on a page you're only visiting, is exactly the shape of mistake that costs a user money. On `develop` `transferFunds` was only reachable from inside `!nonAuthUser` blocks, so this path didn't exist. One attribute fixes it.

### 2. The hero shows a negative AFIT balance on every page load

`wallet.vue:20-40`. The `v-if="!loading && !walletDataRefreshing"` guard is on `.wallet-total` (line 23) only — `.wallet-hero-side` (28-39) is unguarded, and `userTokensWallet` is initialised to `-1` as a sentinel. Observed hero text at initial mount:

```
account_est_val  AFIT_Token -1 AFIT  $-0.04  HIVE POWER 0 HP $0.00
```

`develop` never showed this because all balance display sat inside `<div class="wallet-container" v-if="userTokensWallet >= 0">`. The new hero sits outside that gate. Wrap the whole `.wallet-hero` in the loading guard, or gate on `userTokensWallet >= 0`.

### 3. Fixing the modals exposed wrong "Top HP Holders" numbers

This one isn't your bug, but your fix is what made it visible, and I'd rather it didn't ship as *confidently wrong* data. On `develop` all three modals hit the error branch so nobody ever saw the HP output.

`apiCoinType` forces `'HIVE'` when `balanceType === 'hp_vests'`, so the HP modal fetches liquid HIVE balances and then applies the vests maths (`/1e6 * hivePerVestsRatio`). Run side by side:

```
as shipped     rank1 @cold.dunamu       90.187 HP
forcing VESTS  rank1 @blocktrades  13,340,844.288 HP
```

Wrong accounts, wrong order, and off by roughly 150,000×. The VESTS path works — `coin-type=VESTS&balance-type=balance` returns 200 with 1.19M accounts, and blocktrades' `vesting_shares` converts to ~13.3M HP, matching. Fix is dropping the `if (this.balanceType === 'hp_vests') return 'HIVE'` branch so it returns `this.coinType`, which is already `"VESTS"` from `wallet.vue`. Keep `apiBalanceType` returning `'balance'` — `vests`/`hp`/`vesting_shares` all 400 with `invalid input value for enum`.

A visible error becoming silently wrong money is worse than the original state, which is why this is in the blocking list rather than below.

---

## Non-blocking

- **The column header always reads "Balance"** on all three modals. The component declares `balanceDisplayNameKey`; `wallet.vue` passes `balance-display-key`, so it lands in `$attrs` as a stray DOM attribute and the default wins. Rendered `<th>` is `["Rank","User","Balance"]`. Three renames.
- **`TopHoldersXModal.vue` (AFITX) was missed** — `TopHoldersModal.vue` got the `holders-table` class and the dark-mode contrast rule, its sibling didn't, so AFITX keeps the unreadable dark-mode contrast.
- **The new HIVE history prints raw VESTS.** Verified live on @actifit: `curation_reward` shows `"331.139808 VESTS"` (= 0.205 HP), `producer_reward` `"2793.510145 VESTS"` (= 1.730 HP). The rest of the page converts via `hivePerVestsRatio`; this column doesn't.
- **The 500-op history window is thinner than it looks** — for @actifit that's about 12 hours, 94 financial ops of which 90 are micro-rewards and only 3 are transfers. "Show more" paginates inside that window and can never reach older data.
- **The new empty state is mis-keyed** (`:735`): gated on `tokenMetrics` *and* `tokensOfInterestBal` both being empty, but `tokenMetrics` is the global market list and is non-empty for everyone. So a user with no H-E tokens sees a heading, no rows and no message; the message only appears during an H-E outage.
- **Hero reads `$0` on an H-E outage** while the table below shows real HIVE/HBD. Previously a small toolbar line, now the headline.
- **Duplicated controls** — the old icon toolbar was kept, so four actions appear twice and `account_est_val` renders twice (line 22 and line 80). No regression, just unconsolidated.
- **9 of the 10 new `en_US` keys are English-only.** Not broken — `fallbackLocale: 'en'` means users see English words rather than raw keys — but worth completing. Also some hardcoded English remains: the `Token`/`USD` options, `Ascending`/`Descending`, and several aria-labels.
- **`scrollAction`/`scrollAfitAction` do imperative `insertBefore` on Vue-managed nodes.** They work today (verified: no throw, no duplicates across repeats and a `v-if` toggle), but once DOM order diverges from vnode order that's unsound. `v-if`-driven placement would be safer.
- **RTL:** `ar_AE` is `dir: 'rtl'`, and the new CSS forces `text-align: left !important` on row cells.
- `nuxt.config.js:110` adds a hardcoded `https://api.hive-engine.com/rpc/` fallback for `HIVE_ENG_RPC_NODE`. Defensible hardening, but it's global rather than wallet-only and means a production misconfiguration now silently falls through to a public third-party node instead of failing loudly.
- **No tests**, for ~1200 lines on the highest-stakes page in the app, in a repo that has 309 tests and a `test/unit/pages/*.spec.js` precedent. `normalizeHiveTransaction`, `historyTimestamp`, `sortHistoryEntries` and `displayTokenUsd` are pure and trivial to cover — one pass proved the page is mountable, so the harness cost is already paid.

---

## What you did well, verified rather than assumed

**No balance regression anywhere.** This was the thing I was most worried about in a +1231/−131 rewrite of a money page, so it got checked properly: `renderBalance`, `renderSBDBalance`, `renderSavings`, `renderSBDSavings`, `renderSteemPower`, `vestsToSteemPower`, `hivePerVestsRatio`, `usdVal` and `formattedTotAccountVal` are all logic-identical to `develop`, spot-checked to `"1,234.567 HIVE"` and `"89.12 HBD"`. **No factor-of-1000 error anywhere.** And zero methods or computed properties were removed across the restructure — 12 added, none dropped, so no wallet action was lost.

**The namespacing discipline is genuinely impressive.** In an unscoped page-level style block, where a leak would be easy and invisible, **all 114 new top-level selectors and all 21 new nested media-query selectors sit under `.wallet-*`**. The only non-namespaced nested selector pre-exists on `develop`.

**You applied the earlier CSS scope-leak feedback exactly as suggested** — restored `<style scoped>` and lifted only the namespaced `.dark-mode .holders-table` rule into a separate unscoped block. That's careful reading, not a shortcut.

**You spotted a consequence of your own change and fixed it unprompted:** hiding the `.head-title` row would have killed click-to-sort, so you added the `.wallet-sort-bar` replacement — and the wiring is correct, with the select passing `true` to keep direction and the arrow omitting it to flip.

**And the navbar accident was reverted completely** — `NavbarBrand.vue` isn't in the net diff and its blob hash is byte-identical to `develop`. Messy commit titles, clean final tree. That's what matters.

Also: ~12 new `Array.isArray` guards and initialising `tokenSwapQueue` as `{pendingTransactions: []}` instead of `''` removes a real class of TypeError on slow H-E responses. `historyTimestamp` correctly normalises Hive's naive UTC stamps, and `sortHistoryEntries` doesn't mutate its input — both subtle and easy to get wrong.

Gates: `npm ci` clean, `nuxt build` **exit 0**, lint **exit 0**, and jest is 3 failed / 306 passed — **identical on `develop`**, so zero regressions from this PR.

The three blocking items are all small and in files you're already touching. Fix those and I'll approve and merge, which also clears the `BLOCKED` state.

mcfarhat (approved):
## Approved — all three blockers fixed, and the numbers check out against the chain

Reviewed the **merged** result (this branch + `develop`) at `47342f3e`. Every "fixed" claim below was mutation-proven: I broke each fix and confirmed the check goes red, so we know the verification could have failed.

### 1. Holders endpoint envelope — fixed
`(response.data.holders_result || [])` is in place. Mounted the modal against a live `api.syncad.com` payload: 100 rows render, `error` null, rank 1 `cold.dunamu` shows `179,653,369.798 HIVE` — byte-exact against `condenser_api.get_accounts`. The `|| []` also degrades cleanly to the empty state when the key is missing.

Mutation: `response.data.map()` against that same real payload throws `TypeError`.

### 2. HP modal coin type — fixed, and I checked the value, not just the render
The `if (balanceType === 'hp_vests') return 'HIVE'` branch is gone, so `apiCoinType` correctly returns `VESTS`. Request params verified as `coin-type=VESTS&balance-type=balance`.

I then recomputed the displayed figure independently from `raw/1e6 × (total_vesting_fund_hive / total_vesting_shares)` and cross-checked the **whole top 10 of all three modals** against on-chain balances. HP/VESTS: **10/10 exact**, rank 1 `blocktrades 16,381,027.301 HP`, matching to three decimal places.

Mutation: restoring the `'HIVE'` branch makes rank 1 read `cold.dunamu 111.266 HP` — wrong account, wrong ordering, off by roughly 147,000×.

Worth recording for whoever reads this table next: a couple of HIVE/HBD rows *look* wrong against `get_accounts` and aren't. `steem.dao` is a treasury pass-through — `proposal_pay` credits it hourly and `consolidate_treasury_balance` sweeps it into `hive.fund` three seconds later, so its resting `hbd_balance` is always ~0 while its cumulative flow is in the hundreds of millions. The `null` row is likely the same shape. The data is correct; don't "fix" it by filtering zero-balance accounts, or you'd hide the largest legitimate entry in the table.

### 3. `wallet-quick-actions` guard — fixed
`<nav v-if="!nonAuthUser" class="wallet-quick-actions">`. Mounted with `user: null` on someone else's wallet: nav absent, zero `.wallet-pill`, and — the part that actually matters — **every** wallet-action handler absent from the rendered HTML: `transferFunds`, `powerUp`, `powerDown`, `delegateFunds`, `claimRewards`, `claimSavingsRewards`, `withdrawFunds`, `saveSettings`, `editDelegation`, `updateWalletAddress`, `deleteWalletAddress`. Owner view still renders all five pills.

Mutation: stripping the `v-if` brings all five back.

### 4. The `-1` hero — fixed, and better than what I asked for
I offered "wrap in the loading guard **or** gate on `userTokensWallet >= 0`". You took the second, which is the stronger one: it also covers the Hive-Engine-outage case where `loading` is false but no balance ever arrives. At first paint with pristine `data()` the hero is absent and no `-1` appears. Mutation: without it the hero renders `AFIT_Token -1 AFIT $-0`.

### Merge and gates
`git merge origin/develop` → clean, no conflicts, all 14 `lang/*.js` auto-merged (you appended `off_chain`, #399 appended its description keys, different offsets). **Zero newly-introduced duplicate keys** — I parsed every locale at merge-base, develop, PR head and merged tree and the duplicate sets are identical to develop's. All 14 newly-used `$t` keys exist in `en_US`; nothing renders as a raw key.

- `npm ci` exit 0 · `npm run lint` exit 0 · `npx nuxt build` exit 0
- `npx jest` on this head: **3 failed / 306 passed** — exactly the known `StingChat` ×2 + `disable-email-obfuscation` ×1. **Zero added.**
- `develop` itself is currently at 5 failed — `internal-navigation.spec.js` broke today from an unrelated PR. Not yours, and I'm fixing it separately. Mentioning it so a red `develop` doesn't confuse you.

### Follow-ups — none blocking, will be carded
- **HIVE history prints raw VESTS.** `Curation Reward | 203.521707 VESTS` where the rest of the page speaks HP (that's 0.126 HP). It's *labelled* VESTS so it isn't false, which is why it isn't blocking — but it's a money figure ~1,600× the HP number beside it. `hivePerVestsRatio` is already in scope; roughly a three-line fix.
- `balance-display-key` vs the declared `balanceDisplayNameKey` — silently lands in `$attrs`, so the prop does nothing and leaks onto the DOM. Three call sites. **Pre-exists on develop**, not introduced here.
- `TopHoldersXModal.vue` (AFITX) never got the `holders-table` class or dark-mode rule its sibling has.
- `account_est_val` renders twice — hero plus the old toolbar.
- ~56 pre-existing duplicate keys across the locales, including `en_US.Balance` twice, which these modals depend on. Worth a cleanup ticket.

### What you did well
The final commit is **7 lines across 2 files** doing exactly the three things asked and nothing else — no opportunistic refactoring on a PR that's already +1278, on a page that shows people's money. That restraint is the right instinct and it made this review fast.

Deleting the `apiCoinType` special case rather than adding another branch to it is also the better fix, and the robustness work from the earlier round holds up: I mounted `develop`'s template against your script and it threw `Cannot read properties of undefined (reading 'length')`, while yours rendered clean. The `Array.isArray` guards and the `tokenSwapQueue` initialisation are doing real work.

Merging. Moving the card to Testing — the visual side (action-box layout, expanded power-down, header alignment, dark mode, mobile) can't be checked in jsdom and needs a real browser.

PR #378, opened 2026-08-05
The post screen is hiding activity count and activity type

ShahdQaddoura:
Screenshot (739)
Screenshot (740)
Screenshot (742)
Screenshot (741)
Updates report activity labels to use the shared --text-color theme variable, ensuring correct light/dark mode colors without extra overrides.

### DIFF
--- a/components/ReportModal.vue
+++ b/components/ReportModal.vue
@@ -829,6 +829,13 @@ export default {
   text-align: center;
 }
 
+html:not(.dark-mode) #reportModal .modal-body img[src*="ACTIVITYCOUNT"] + .text-center,
+html:not(.dark-mode) #reportModal .modal-body img[src*="ACTIVITYCOUNT"] + .text-center *,
+html:not(.dark-mode) #reportModal .modal-body img[src*="ACTIVITYTYPE"] + .text-center,
+html:not(.dark-mode) #reportModal .modal-body img[src*="ACTIVITYTYPE"] + .text-center * {
+  color: #000 !important;
+}
+
 .modal-author {
   margin-left: 10px !important;
 }
--- a/components/ReportModal.vue
+++ b/components/ReportModal.vue
@@ -805,7 +805,7 @@ export default {
   display: flex;
   align-items: center;
   justify-content: center;
-  color: #fff !important;
+  color: var(--text-color) !important;
   text-align: center;
 }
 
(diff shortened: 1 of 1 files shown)


mcfarhat (commented):
Reviewed — the fix is correct and low-risk. Root cause: the existing rules force `color:#fff !important` on the activity fields unconditionally, so in **light mode** that white text is invisible against the light modal. Your override fixes it well:

- ✅ The `<style>` block isn't scoped, so the `html:not(.dark-mode)` selector applies as intended.
- ✅ Verified `dark-mode` is toggled on `document.documentElement` (`<html>`) in `plugins/dark-mode.js`, so `html:not(.dark-mode)` correctly targets **light mode only** — dark mode still gets white. No leak.
- ✅ The extra `html:not(...)` raises specificity above the base `#reportModal ...` rule, so the `#000` override wins.

### 🟡 One thing to confirm before merge
The existing white-text rule (around line 817) also styles **`ACTIVITYDATE`**, but your override only covers `ACTIVITYCOUNT` and `ACTIVITYTYPE`. If the activity **date** also renders in the modal, it'll still be invisible in light mode. Please either **add `ACTIVITYDATE` to the light-mode override** too (for consistency), or confirm it's intentionally excluded / not displayed.

### 🟢 Optional (non-blocking)
A cleaner long-term approach would be to gate the *original* `#fff` rules under `.dark-mode` rather than layering a `:not()` override — but the current fix is fine to ship.

Looks good to approve once the ACTIVITYDATE point is confirmed. 🙏

mcfarhat (changes requested):
Formally marking changes-requested to reflect status (see my detailed review above): please add `ACTIVITYDATE` to the light-mode override for consistency, or confirm the date isn't rendered in the modal. Otherwise the fix is correct.

mcfarhat (approved):
✅ Approved. Verified: base color changed to `var(--text-color)` (#000 light / #fff dark), so Activity Count/Type/Date are readable in both themes incl. ACTIVITYDATE. Lint clean, confirmed visually. 🎉

mcfarhat (approved):
✅ Approved. Verified: base color changed to `var(--text-color)` (#000 light / #fff dark), so Activity Count/Type/Date read correctly in both themes incl. ACTIVITYDATE. Lint clean, confirmed visually. 🎉

PR #382, opened 2026-08-10
Unify the post bottom strip

reemnii:
Unified the post bottom strip across PostModal.vue, ReportModal.vue, and pages/_username/_permlink/index.vue using CardActions.vue


mcfarhat (changes requested):
Great architecture overall — unified strip in a **scoped** `CardActions.vue`, all surfaces delegate to it, feed cards inherit it, tests/lint green. But there's a **crash blocker** in the report modal.

## 🔴 Report modal crashes for logged-out users
Repro: open `/activity` (logged out) → click any report → full-page error **"Cannot read properties of undefined (reading '$el')"**, blank modal.

**Root cause** (a regression this PR introduced):
- The old ReportModal vote button was guarded: `v-if="this.$parent.user && userVotedThisPost() == true"` — so `userVotedThisPost()` was **never called when logged out**.
- This PR replaced it with `CardActions` and binds `:hasVoted="userVotedThisPost()"` (`components/ReportModal.vue:36`) — now called **unconditionally on every render**.
- `ReportModal`'s `userVotedThisPost()` (line 615) is **not null-safe**:
  ```js
  userVotedThisPost() {
    let curUser = this.user.account.name;   // this.user is null when logged out -> throws
    ...
  }
  ```
- The throw aborts the modal-body render, so `$refs.remarkableContent` never mounts, and `attachImageErrorHandlers()` (line 693, `vm.$refs.remarkableContent.$el`) then dies on the missing ref — that's the `$el` message you see.

**Fix** — guard the method (mirror what the permlink page already does at line 527):
```js
userVotedThisPost() {
  if (!this.user || !this.report) return false;
  let curUser = this.user.account.name;
  this.postUpvoted = this.report.active_votes.filter(voter => voter.voter === curUser).length > 0;
  return this.postUpvoted;
}
```
Nice-to-have defense-in-depth: also null-check the ref at line 693 (`const rc = vm.$refs.remarkableContent; if (!rc || !rc.$el) return;`).

## ✅ Scope (for your reassurance)
- **PostModal** — safe: its binding is guarded (`:hasVoted="user && userVotedThisPost() == true"`).
- **Permlink page** — safe: its `userVotedThisPost()` already has `if (!this.user || !this.report) return false;`.
- Only **ReportModal** is affected. Logged-in users don't hit it, which is why the modal-post test earlier looked fine.

Everything else in the PR is good — this is the one thing blocking merge. Please also give the report modal a quick logged-out pass after the fix. 🙏

mcfarhat (changes requested):
Reviewed at `71bdb93a`. Two independent passes, both running the real Nuxt app in headless Chrome and A/B-ing it against `develop` — plus a third checkout of the pre-fix parent so the original crash could be reproduced rather than assumed.

## ✅ The blocker I raised is genuinely fixed

Reproduced on the parent commit `5f81e14`, verbatim:
```
Cannot read properties of null (reading 'account')   at userVotedThisPost
Cannot read properties of undefined (reading '$el')  at ReportModal
```
— modal gone from the DOM entirely. At your head, logged out: modal opens, body renders, footer present, reblog correctly hidden, **0 console errors**.

I mutation-tested it, and you shipped belt-and-braces — worth knowing which piece is doing the work:

| removed | result |
|---|---|
| the method guard only | no crash — the ternary at `ReportModal.vue:57` short-circuits first |
| the ternary only | no crash — the method guard catches it |
| **both** | crash returns |

So either guard alone is sufficient. Your `$refs.remarkableContent` null-check is right too — with both value guards removed, only the *first* error fired, because that check absorbed the `$el` cascade. And your new `ReportModal.spec.js` mounts with `user = null` and asserts the method doesn't throw: a real regression test.

Also good: logged-in paths all verified against real accounts — voter shows `hasVoted: true` with the active class, non-voter false, author correctly loses reblog, vote modal opens in all three.

Two things block merge, and one of them concerns me more than the bug itself.

---

## 🔴 1. There is no working way to reply to a post from its own page

`pages/_username/_permlink/index.vue:84` binds `@reply="toggleCommentBox"` — and **that method doesn't exist on the page component**. The old control used the inline `commentBoxOpen = !commentBoxOpen`; consolidation replaced it with an event nobody listens for. Every load of `/@author/permlink` logs:

```
[Vue warn]: Property or method "toggleCommentBox" is not defined on the instance but referenced during render
[Vue warn]: Invalid handler for event "reply": got undefined  ---> <CardActions>
```

Clicking it does nothing — measured `replyBoxOpen: false` before and after, in every identity state (anonymous, logged-in non-author, already-voted, the post's own author). And because `index.vue:857` sets `.report-head .legacy-post-actions { display: none }`, the old reply affordance is hidden, so there is no fallback. `PostModal` is unaffected — it does have the method.

**The part I want you to take away:** your test suite is green *because it patches this in*. `test/unit/pages/.../index.spec.js:160` does `ReportPage.methods.toggleCommentBox = jest.fn()` in `beforeEach` and deletes it in `afterEach`. Adding a missing method to the component under test makes the test pass and the feature stay broken — permanently, and invisibly. When a test needs a method the component doesn't have, that's the test telling you something. Please fix the component (`@reply="commentBoxOpen = !commentBoxOpen"`, or add the method) and drop the patch so the spec exercises the real thing.

## 🔴 2. Post-body images now overflow the column

`index.vue:844` changed `<style>` to `<style scoped>`. Inside that block is `img { max-width: 100%; }` — which was the **only** constraint on markdown images on that page (the global rule is `.modal-content .modal-body img`, and `SafeRemarkable` renders as `col-md-12`, outside any modal).

Scoped styles compile to `img[data-v-16336889]`, and markdown-generated `` elements carry no `data-v-` attribute, so the rule is now dead. Measured on the same post at 1280px, 822px content column:

| | develop | this PR |
|---|---|---|
| computed `max-width` | `100%` | **`none`** |
| rendered width (1280px natural) | 792px | **1280px** |

Images run off the right edge of the column. This is the classic Vue 2 scoped-style trap: the hash is appended to the **last compound selector**, so any rule targeting runtime-generated DOM (`v-html`, remarkable output) silently stops applying. Either drop `scoped` here, or use `::v-deep img { max-width: 100% }`.

Related, same cause and currently harmless: in `ReportModal.vue:811-814` four light-mode activity-card overrides didn't get `::v-deep` while their siblings did, so they compile dead. No visible effect today only because a surviving rule sets the same colour — but it'll break the moment `--text-color` changes.

---

## Regressions worth fixing in the same pass

**Feed cards grew a row.** `CardActions.vue:41` gives `.card-actions` `flex: 1 1 auto` but no `display: flex`, and `.post-detail-footer__extra-actions` relies on `margin-left: auto` — a no-op outside a flex container. In the three modal/page surfaces the parent *is* flex so it looks right; in `Post.vue`/`Report.vue` the parent `.post-footer` is a plain block, so the children stack. Measured on a blog feed: footer 66px → **89px**, share/edit/read-more dropping from the action row to their own line. That's every card in explore, search, blog, comments, tag, activity and community listings.

**The AFIT chip disappeared from two of three payout strips.** `{{ afitReward }} AFIT_Token` was removed from both the report modal (`develop:ReportModal.vue:86-87`) and the post detail footer (`develop:index.vue:96-97`). PostModal never had it, so unifying levelled *down* to the weakest surface. Live footer went from `"31 1 0 AFIT 0.290 HBD 1"` to `"31 1 0.290 HBD 1"`. `afitReward` is still fetched on every load and now has zero template consumers — dead state plus a wasted request. Was dropping Actifit's own token from the payout display intentional? I'd assume not.

**`$safeUrl()` dropped** at `ReportModal.vue:21`. Inert here since author/permlink are chain-constrained, but `plugins/sanitize.js` exists specifically to force unknown schemes to `#`, and silently removing a hardening helper in a layout refactor is the kind of thing that's hard to notice later. Please restore.

**`hasVoted` prop-type warning.** `PostModal.vue:83` passes `user && userVotedThisPost() == true`, which is `null` when logged out, against `{ type: Boolean, required: true }` — so every logged-out post modal logs `Invalid prop: type check failed for "hasVoted". Expected Boolean, got Null`. One-liner: `!!(user && userVotedThisPost())`. Your three call sites currently disagree on this contract; `ReportModal.vue:57` has it right.

---

## Credit where it's due

**Tests: 313 pass / 3 fail vs develop's 306 / 3** — the same three failures by name on both sides (`StingChat` ×2, `disable-email-obfuscation`), all pre-existing and not yours. Lint clean. Net **+7 tests**, and you rewrote `PostModal.spec.js` from four `readFileSync`-and-string-match assertions into three real `mount()` tests. Fewer tests, materially better ones — that's the right trade and I'd like to see more of it.

**Locales:** `git diff develop...pr382 -- lang/` is empty; no new strings. I still loaded all 14 `lang/*.js` and checked every `$t()` key this PR touches — `votes`, `comments`, `Reply`, `reblog`, `copy_link`, `author_payout`, `voters_payout`, `hive_payouts_wait`, `more_token_rewards`, `AFIT_Token` — present with real translations in all 14.

**Comment threads are clean.** I drove 40 comment cards and voting always targeted the exact comment, never the parent post. Feed cards at scale are clean too — 20 cards, all counts distinct and correct, modal matched its card, no state bleed.

One scope note: despite the title, comments and posts aren't actually unified — `Comments.vue` isn't in the diff and keeps its own strip with edit/delete and a replies toggle, while `CardActions` has reblog and no edit/delete. This restyles the three *post* strips to resemble the comment strip. Worth saying so in the description.

**Not yours, separate ticket:** `translate_content` / `translating_content` are missing from all 14 locales — the code calls `$t('translate_content', 'Translate Content')` expecting the second argument to be a fallback, but in vue-i18n v8 it's the *locale*. Live console confirms it. Same pattern on develop in four components.

The crash fix is solid work and the test rewrite is a real improvement. Ping me when the reply handler and the scoped-style issue are in.

mcfarhat (approved):
## Approved at `9f6fc8d7` — all six fixed, and the test lesson landed

Every item verified by execution and mutation-tested. Gates clean: `npm ci` 0, lint 0, `nuxt build` 0, and jest **315 passed / 3 failed** against develop's **306 / 3** — same three known failures by name, so **+9 tests and no regression**.

### The two blockers

**1. `toggleCommentBox` — fixed, and this is the part I most wanted to see.** The method exists at `index.vue:416`, but more importantly **you deleted the `jest.fn()` patch**. `grep -rn "toggleCommentBox\|methods\.[a-zA-Z]* = " test/` returns nothing, and you replaced it with a mount-and-click test that drives the real component with `CardActions: false` so the real child renders.

I mutation-tested it: deleted the method from the page, re-ran your spec, and got `✕ toggles the reply editor through the page method — Expected: true, Received: false`. **The test now fails when the bug is present.** That was the whole point of the feedback and you got it exactly.

**2. Image overflow — fixed, and fixed better than develop had it.** Wrapping `SafeRemarkable` in `.post-body` and using `.post-body ::v-deep img` puts the scope attribute on an *authored ancestor*, so the compiled rule is `.post-body[data-v-1780c93c] img{...}` — a bare `img` descendant that matches markdown-generated nodes.

Measured `getComputedStyle().maxWidth` against the real compiled CSS, with a genuine `SafeRemarkable` render of `alt`:

```
markdown img in post body            100%
same, ::v-deep removed (mutation)    ""     dead
same, .post-body wrapper removed      ""     dead
```

Both mutations go red. And this is strictly better than develop's global `img { max-width: 100% }`, which leaked to every page in the app.

### The other four

`.card-actions` now has `display: flex` (computed `flex`; strip it and the two-row wrap returns). The AFIT chip is restored verbatim on both surfaces — I diffed against develop and mounted with `afitReward: 42` to confirm it renders. `$safeUrl()` is back on `ReportModal.vue:25` matching develop, with the spec mock added so the mount exercises it. And `hasVoted` is now `!!(user && …)` with your own test asserting `false` when logged out — zero `Invalid prop` warnings across all surfaces in all three identity states. The light-mode `::v-deep` group compiles correctly too.

### Your new commit, reviewed as new code

Extending `CardActions` to absorb the comment strip is the unification the PR title always promised — I'd noted earlier that comments weren't actually unified, and now they are. It deletes ~45 lines of duplicated CSS, and bridging the theme with `--post-footer-muted: var(--cmt-muted)` on `.cmt-footer` is genuinely elegant: custom properties inherit through the DOM regardless of scope, and `.cmt-thread` is always an ancestor, so they resolve.

I mounted `Comments.vue` with the real `CardActions` and drove every handler. All fire correctly — edit, delete (including a clean abort on declining `confirm()`), reply, replies-toggle with `aria-expanded`, and vote committing the **comment** rather than the parent post. Visibility gating is right in all three identity states, and `children: 0` correctly degrades to the static span. **Zero Vue warnings.** No handler is bound to a missing method — the original blocker's bug class does not recur.

Branch is **0 commits behind** develop and trial-merges cleanly. #399's `onchainDesc` fallback is byte-identical; no locale changes at all.

---

## One thing worth a quick follow-up commit

**Comment-body images lost a constraint develop has.** Measured: `.cmt-body img` computes `max-width: 100%` on develop via that leaky global rule, and **empty** on this branch. It dates to your earlier `<style scoped>` change rather than the new commit, and item 2 only restored the *post* body. `PostModal`/`ReportModal` are still covered by `main.sass`, but comment images on `/@author/permlink` can now overflow their card.

One line in Comments.vue's unscoped block: `.cmt-body img { max-width: 100%; height: auto; }`. I'm merging rather than holding for it since it's cosmetic and you're moving fast, but it's the first thing I'd add.

## Smaller notes

- **`pages/login.vue` is in this PR** — a null-guard on the HiveAuth QR `$refs`. The code is correct, but it has nothing to do with the post strip, isn't in the PR description, and is buried under "Refactored styles" in the commit message. Worth its own PR next time.
- ~40 lines of dead scoped CSS remain at `index.vue:896-929`, targeting nodes that now carry CardActions' scope id. Harmless, and you *did* clean the equivalent blocks out of both modals — worth finishing for consistency.
- The replies toggle went from `<button>` to `<a href="#">` and lost its `:focus-visible` outline with no replacement in `CardActions`. `aria-expanded` survives, but screen readers now announce "link" and keyboard users lose the focus ring.
- `Comments.spec.js` is still `readFileSync` string-matching, so the Comments↔CardActions integration — the largest behavioural change here — has no mount test. Everything works; it's just guarded by the weaker pattern you deliberately moved away from in `PostModal.spec.js`.

## Not verified

Real-browser layout. jsdom computes the cascade but does no layout, so item 3 is proven at the level the diagnosis was made (`display` computes to `flex`, `margin-left` to `auto`) rather than by measuring rendered pixels. Same for actual image overflow at a given viewport.

---

Merging. Moving the card to Testing so the visual side gets a real browser — feed-card row height, post-body images at a narrow viewport, and the comment strip in both light and dark mode.

The turnaround here was the good part: two blockers fixed at the root cause rather than papered over, and on the test you deleted the crutch instead of working around it.

PR #395, opened 2026-08-10
fix(i18n): restore account-recovery locale keys, drop level_short leading dash

mcfarhat:
Two locale regressions currently live on `develop`, both user-visible. Follow-up to #364.

## 1. `invalid_account_name` / `account_lookup_failed` missing from all 14 locales

`1c4f0d3` ("fix market character encoding") restored `lang/` from a `develop` snapshot that predated **`999db8a` (#363, "Fix invalid username loading (account recovery)")**, so both keys added by that PR were silently dropped from every language file.

They are still referenced:

- `pages/password.vue:538` — `this.errorFetch = this.$t('invalid_account_name')`
- `pages/password.vue:569` — `this.errorFetch = this.$t('account_lookup_failed')`

So password recovery has been rendering the literal strings `invalid_account_name` and `account_lookup_failed` to users in all 14 languages since #364 merged.

Restored verbatim from `999db8a`, preserving the original translations rather than re-translating — inserted after the `user_not_found_error` anchor they originally followed.

## 2. `level_short` kept a leading `" - "` in 7 of 14 locales

`es_ES`/`fr_FR`/`pt_PT` (`" - N"`), `ru_RU` (`" - Ур."`), `uk_UA` (`" - Рв."`), `tr_TR` (`" - Sv"`), `hi_IN` (`" - एल"`).

Rendered in the level pill as red text on a pale background, `- N3` reads as a negative number. The other 7 locales were already corrected; this brings the remaining 7 into line.

## Verification

- All 14 files `require()` cleanly, with both keys present and no `level_short` leading dash — asserted programmatically, not by inspection. Cyrillic and Devanagari values verified intact (no encoding damage).
- `npx jest`: **306 passing / 3 failing**, which is **byte-identical to `develop`'s baseline** — I ran the suite with and without this change. `StingChat.spec.js` and `disable-email-obfuscation.spec.js` fail on `develop` too and are unrelated.
- Diff is 14 lang files, +35/−7. No other files touched.

## Note

Prepared automatically as a follow-up to the #364 review. Neither regression was introduced deliberately — #1 is collateral from an encoding recovery, and #2 was a partial fix where a grep pattern only matched the Latin-script locales (my own miss in an earlier review round).


mcfarhat (commented):
✅ Reviewed — clean, correct, well-scoped. (Can't click Approve since it's my own PR, but this is good to merge.)

Verified independently:
- **Regression confirmed on develop**: `invalid_account_name` had 0 occurrences (referenced at `pages/password.vue:538/569` → users saw the raw key in password recovery), and ru_RU's `level_short` carried the leading dash.
- **Fix confirmed**: both keys present in **14/14** locales; **no** `level_short` leading dash anywhere; cleaned values sensible (`Ур.`/`Рв.`/`Sv`/`एल`/`N`).
- Restored non-Latin values intact (Cyrillic/Arabic/CJK/Devanagari — no encoding damage).
- Diff **lang-only** (14 files, +35/−7, 0 others); all 14 `require()` cleanly; locale tests 43/43 pass; `npm ci` check green.

PR #396, opened 2026-08-11
Feature/revamp market screen

reemnii:
Made the sidebar sections collapsible and added lock icon for "Locked" products.


mcfarhat (changes requested):
Reviewed at `95f12fb7`. Two independent passes, both driving the real Nuxt app in headless Chrome, plus one check I ran myself to settle a disagreement between them.

**The accordion itself is well built, and I want to be specific about why**, because these are the things that usually go wrong and didn't:

- **Collapse state is keyed by `group.type`**, a stable identifier — not an array index, not the translated label. So changing locale can't corrupt it, and a group disappearing and coming back reattaches correctly. Verified: collapse "Physical Products" → apply a filter that removes the group → clear it → state comes back on the right group.
- **It's a real `<button type="button">`**, not a `div` with `@click`. Enter and Space both work, `aria-expanded` updates, and because you used `v-if` the panel leaves the DOM entirely — so a collapsed group has exactly **one** focusable element and Tab goes to the next header. No hidden-but-tabbable controls.
- **Nothing entitlement-related is affected.** Every purchase/activate control lives in the detail panel, not the sidebar (measured: 0 in the sidebar, 9 in the detail panel), and the lock icon is `aria-hidden` decoration — nothing derives state from it.
- **No regressions to #364**: sort arrows render as real glyphs with no double-encoding, search autofill attributes intact, "Locked" rename holding, saved-products panel still renders, and no horizontal overflow at 375px in any collapse state.
- **Jest 306 pass / 3 fail — identical to `develop`**, and ESLint clean. The 3 failures are pre-existing and not yours.
- **No new user-facing strings**, confirmed by loading all 14 `lang/` files and comparing the `$t()` call sets rather than grepping. `market_state_unavailable` ("Locked") is genuinely translated in all 14.

One thing blocks merge.

---

### 🔴 Search results silently vanish inside collapsed groups

`groupedProducts` derives from `filteredProducts`, but nothing expands a group when a search matches inside it — `searchQuery` only feeds `ensureSelection`. Combined with `v-if`, matches inside a collapsed group are removed from the DOM while the counts still report them.

I reproduced this against the live catalog:

```
all groups collapsed + search "Resistance"
  results label : "6 Products"          group badge  : "PHYSICAL PRODUCTS 6"
  rendered rows : 0                     sidebar text : "PHYSICAL PRODUCTS 6"

control — same search, groups expanded
  rendered rows : 6                     sidebar text : "...Resistance Band AFIT 196,719.27 AFIT..."
```

So the screen says "6 Products" in two places and shows none. The other pass hit it a different way — collapse "Game", search `Water Bottle` → toolbar "5 Products", badge "5", zero rows, large empty panel.

**On desktop it's partly masked**, because `ensureSelection` auto-selects the first match, so the detail panel shows match #1 — but matches #2–6 are unreachable. **On mobile (390×844) there's no masking at all**: `mobileShowDetail` is false on load, so the sidebar is the entire screen. The user searches, is told there are matches, and sees a blank list with no explanation.

**Fix:** when `searchQuery` is non-empty (or `activeFilterCount > 0`), force groups open — or render an inline "N matches hidden — expand" affordance on the collapsed header. Auto-expanding on search is the more conventional behaviour and probably what people expect.

The same reasoning applies to the filter drawer, so please handle both together.

---

### Worth fixing in the same pass

**Scroll jump when collapsing a group above the viewport.** With the last header anchored at y=400, collapsing the first group moved it to **y=−1154** — a 1554px jump, doc height 6145→3252. The content you were looking at is flung off-screen. Browser scroll anchoring doesn't rescue it because the height animates over 280ms. Worth compensating scroll position, or setting `overflow-anchor`.

**The collapse transition clips tall groups.** `.sidebar-collapse-enter-to/leave` caps `max-height: 1400px`, but the Game group is taller than that in real data — one pass measured 2894px against the live catalog, the other 1483px against fixtures. Either way it animates up to the cap, then the class detaches and height snaps to full, so the last stretch pops instead of sliding. It'll get worse as the catalog grows. A JS-driven height (or a much larger cap) fixes it.

**`aria-controls` missing (minor).** The toggle has `aria-expanded` but the panel has no `id`/`role="region"` to point at, so a screen reader can't jump from header to panel.

---

### Not yours — flagging so nobody attributes them to this PR

- **`/ar/market` never sets `dir="rtl"` on `<html>`** despite the RTL locale being configured in `nuxt.config.js:350`. Repo-wide and pre-existing. Your header mirrors correctly once direction is forced, so the accordion is fine — but the hardcoded `text-align: left` on `.sidebar-group-toggle` is a latent RTL issue if that title ever wraps.
- `fetchGadgetPrizeCycle` (`market.vue:616-620`) calls `reject` with no enclosing Promise → `ReferenceError` whenever the response shape is unexpected. Same pattern at 543-544 and 558-559.
- `components/Product.vue:13` does `this.product.image.startsWith('http')` unguarded and hard-crashes when `image` is absent — notably, your own code at `market.vue:787` guards this correctly.

I'll raise those separately.

Nice work — the hard parts of an accordion are the ones you got right. Ping me when the search behaviour is in and I'll re-check quickly.

mcfarhat (changes requested):
Re-reviewed at `3a34d0d0`, driving the real app against the live catalog in headless Chrome at 1440×900, 390×844 and 375×812.

## The blocker is properly closed

Both of my original scenarios, at both viewports:

```
                                      results   badge   rows rendered
all collapsed + "Resistance" @1440      6         6        6   (was 0)
all collapsed + "Resistance" @390       6         6        6   (was 0)
collapse Game + "Water Bottle" @1440    5         5        5   (was 0)
same @390                               5         5        5   (was 0)
```

And the approach is the right one — it auto-expands **without mutating** the saved state, so clearing the search restores exactly what the user had collapsed. Verified across a full cycle: `{"ingame":false,"ebook":false,"real":false}` before, during and after. The filter drawer behaves the same way (52 rows with a currency filter, badges summing to the total, state untouched). You also forced `mobileShowDetail=false` while searching, which fixes the blank-list-on-mobile half of the report. Good, complete fix.

Everything I'd previously verified still holds: collapse state keyed by the stable `group.type` (checked under `zh`, where the labels become 游戏/电子书/实体产品 and the keys stay `ingame/ebook/real`), one focusable element per collapsed group with Tab landing on the next header, no purchase control anywhere in the sidebar (63 buttons = 60 rows + 3 toggles), all the #364 fixes intact, no horizontal overflow at 375px in any state, and jest **306 pass / 3 fail — identical to develop**, lint clean.

Two things to fix, one of which the fix introduced.

---

### 🔴 New: the collapse toggle is silently inert *and latches* while a filter is active

With a currency filter on, clicking a group header three times:

```
filter HIVE active:  aria-expanded=true  rows=52  state={"ingame":true,...}
after click #1:      aria-expanded=true  rows=52  state={"ingame":false,...}   <- recorded, invisible
after click #2:      aria-expanded=true  rows=52  state={"ingame":false,...}
after click #3:      aria-expanded=true  rows=52  state={"ingame":false,...}
filter removed:      aria-expanded=false rows=15                              <- snaps shut
```

Cause is at `pages/market.vue:743-744`:

```js
const isOpen = this.isSidebarGroupOpen(groupType);   // returns true unconditionally while filtering
const nextState = !isOpen;                            // therefore always false
```

`nextState` is derived from the short-circuited *getter* rather than the persisted value, so every click computes `false` and further clicks can't undo it. Nothing appears to happen, then the group snaps shut later from a click the user made minutes ago.

A search is transient, so this barely shows there — but a **filter is persistent state**, which means a user browsing with a filter on can't collapse anything at all, and gets a surprise when they clear it. Fix is one line: read `this.sidebarGroupState[groupType]` directly rather than calling `isSidebarGroupOpen()`.

### 🔴 Still open: the scroll jump — the compensation code cannot execute

Measured at 375×812 with the first header 2609px above the viewport:

```
last header y: 400 -> -1390   (moved -1790px)
doc height 6149 -> 3255
```

Same defect as before, slightly larger. The compensation you added at `:747-756` can never run, and it's a Vue 2 gotcha rather than a logic error: **`:ref` inside a `v-for` gives you an array**, not an element. Your refs are inside the `v-for="group in groupedProducts"` at `:189`, so:

```js
typeof toggleEl.getBoundingClientRect === 'function'   // -> false, on an array
```

`shouldCompensate` is therefore always `false` and `window.scrollBy` never fires. I dumped `$refs` live to confirm — `isArray: true`, `hasGetBoundingClientRect: false` for all six refs. `this.$refs[...][0]` fixes it.

Two related notes: the three new `overflow-anchor: none` rules make no measurable difference (I forced them back to `auto` and got **−1790px either way**), and on desktop `window.scrollBy` is the wrong scroller anyway, since `.sidebar-scroll` scrolls independently.

---

### Cheap cleanups worth folding in

- **The transition hooks are dead code.** The `max-height: 1400px` cap is gone, so the clipping is fixed — but the JS hooks meant to replace it are bound on the child `<div>` at `:215-220` instead of on `<transition>` at `:207`, so Vue registers them as DOM listeners for events that are never dispatched. I instrumented all six and ran a full collapse/expand: **zero hooks fired**, `transitionDuration: 0s`, no inline height ever set. So collapse/expand is now instant with no animation. Either wire them to `<transition>` or delete the ~55 lines.
- **`aria-controls` points at an id that doesn't exist when collapsed.** The attribute and the panel's `id`/`role="region"`/`aria-labelledby` are all there now — but the panel is `v-if`, so while collapsed `aria-controls` is a dangling IDREF, measured `targetInDom: false` on all three groups. That's invalid ARIA in exactly the state where the relationship matters. Either `v-show`, or drop the attribute when collapsed.

### Housekeeping

The PR is **BEHIND `develop`** by 4 commits (the #397 cookie-banner merge and follow-ups). They only touch `Footer.vue` and `VoteProposalModal.vue`, disjoint from `market.vue` — I ran a trial merge and it's **clean, zero conflicts**. Just needs a merge or rebase.

One thing that looks alarming but isn't: your new commit was authored before the develop merge and merged back afterwards, so a naive `git diff 95f12fb7..6b0f64bd` appears to delete workflow and lang files. That's a fork-point artifact — the real net change against the merge base is **`pages/market.vue` only, +264/−41, one file**. No collateral.

The search fix is genuinely good and it handles the cases I couldn't have listed for you in advance (filter drawer, logged-in grouping, mobile detail panel). Fix the toggle derivation and the `$refs[0]` and this is done.

mcfarhat (changes requested):
## Re-reviewed at `66c012d2` — all four previous items fixed, one new blocker

Every item from 11 Aug is genuinely closed, and each survived mutation testing — I broke the fix and confirmed the check goes red, so none of these are vacuous passes.

**1. Toggle inert and latching under a filter — fixed.** Filter HIVE on, three clicks → `aria-expanded` false/true/false, rows 5→2→5→2.

**2. Scroll compensation could never execute — fixed, and you went past what I asked.** `getVueRefElement` unwraps the `:ref`-in-`v-for` array; `scrollBy({top:-512})` fires and the inner `.sidebar-scroll` path moves `scrollTop` 900→567. You also pick the inner scroller when it's the real one and compare against the *scroller's* top rather than 0 — which answers the "wrong scroller on desktop" note I raised separately.

**3. Transition hooks were dead code — fixed.** Moved onto `<transition>`; hooks fire, inline `height: 0px`, correct leave classes, and a rapid double-toggle mid-animation recovers cleanly.

**4. `aria-controls` dangling IDREF — fixed** via `v-if`→`v-show`. The panel resolves by id while collapsed, `role=region`, `display:none` so there are still no hidden tabbables.

That `v-if`→`v-show` swap is one elegant move that fixed the dangling IDREF *and* enabled the height animation without introducing hidden tabbables. And items 2 and 3 are the two genuinely subtle Vue 2 traps — `:ref` inside `v-for` yielding an array, and hooks belonging on `<transition>` rather than the child. Both nailed.

**A method note, because it nearly cost you an unfair finding.** My first pass showed *zero* transition hooks firing and I was about to report item 3 as still dead. `@vue/test-utils@1.3.6` stubs `<transition>` by default. I built a control component of identical shape, proved the stub was swallowing the hooks, un-stubbed and re-ran. Your code is correct — the harness was lying.

---

## Blocking — the filter fix got reverted

`66c012d2` deletes `|| this.activeFilterCount > 0` from `shouldAutoExpandSidebarGroups`, which reverts the filter half of the fix I signed off at `3a34d0d0`. So matches are hidden inside collapsed groups again:

```
collapse all three groups, apply currency filter HIVE
  toolbar "5 Products"   badges 3, 2   rows in DOM 5   VISIBLE 0

control (same state, search "Gadget"):  toolbar 6, VISIBLE 6
milder: collapse only Physical Products + HIVE → toolbar 5, VISIBLE 3 (2 hidden)
```

**Your reasoning was sound, and it was my complaint that prompted it** — a persistent filter forcing groups open is exactly why the toggle felt inert. But these two goals shouldn't be traded against each other; I asked for search and filter to be handled together and this satisfies one by giving up the other.

The shape that gets both: drop the computed override, and on `shouldAutoExpand` going false→true snapshot `sidebarGroupState` and open all, on true→false restore the snapshot. Then the toggle is always live *and* matches are never hidden. Roughly ten lines.

## Coupled side effect you probably didn't see

`shouldAutoExpandSidebarGroups` has a **second consumer** — `ensureSelection` at `market.vue:720`, which forces `mobileShowDetail = false`. Narrowing the computed to search-only means **applying a filter on mobile no longer returns you to the list**: `mobileShowDetail` stays `true`, where it was `false` at `3a34d0d0`. And since `ensureSelection` may reselect `filteredProducts[0]`, you can land on the detail page of a product you never chose.

Easy to miss — the computed reads like it only drives group expansion.

## Non-blocking

- The toggle is still inert (no longer latching) **during a search**: three clicks leave `aria-expanded=true` and rows unchanged while the underlying state alternates, then clearing the search snaps the group shut. The same watcher fix resolves it.
- `v-show` keeps every collapsed group's rows mounted, so collapsing no longer reduces DOM size. Fine at 60 products; worth knowing if the catalogue grows.

## Verified clean

**Money.** All 15 sidebar rows' AFIT/HIVE figures match `getProductAfitPrice`/`getProductHivePrice` exactly — **zero mismatches** — and those reproduce `Product.vue`'s own `getPrice()` formula (detail 33 HIVE == sidebar 33 HIVE). HIVE-only products render an em dash for AFIT rather than `0`. With a zeroed price feed every price is `—`; no `NaN`, `Infinity` or `undefined` anywhere.

**Logged-out.** The sidebar contains only 3 group toggles and 15 row buttons — no anchors, inputs, or buy/activate/checkout controls, no `bought`/`activated` groups, status fieldset is `v-if="user"`. No repeat of the #377 leak.

**Group state is keyed by identity, not index** — I verified by prepending a group to shift indices, and mutating to index-keying goes red. That's the classic "wrong row expands when the list is filtered" bug and you avoided it.

i18n: all 38 reachable keys exist in all 14 locales.

## Gates

```
npm ci exit 0    lint exit 0    nuxt build exit 0
jest  306 pass / 3 fail    develop baseline (same clone): 306 / 3 — identical
```

Zero added failures. You merged develop yourself, so the trial merge is a no-op with zero conflicts — `BLOCKED` is just the review gate.

No new tests were added, which is why I keep having to rebuild this harness each round.

## Not verified

Real-browser layout: whether the 280 ms animated collapse combined with the *immediate* `scrollTop` correction produces a visible jump — the correction now lands ~280 ms before the content actually moves. jsdom has no layout engine, so that needs a headless-Chrome pass or your eyes.

Pre-existing and not yours: `reject is not defined` at `market.vue:678` in `fetchGadgetPrizeCycle` — it hard-crashed my node process during testing, so it's real and live on develop.

---

Fix the auto-expand so it does both, check the `ensureSelection` coupling, and this is done. The four items you were asked for are properly closed.

mcfarhat (approved):
## Re-reviewed at `d0a85a84` — both asks closed. Approving and merging.

Last round I asked for two things: fix the auto-expand so the toggle stays live *and* matches are never hidden, and check the `ensureSelection` coupling. Both are done, and done the right way.

### The auto-expand fix — verified by mutation, not by reading

You dropped the computed override (`isSidebarGroupOpen` is now a plain state read) and moved the behaviour into a watcher that snapshots on false→true and restores on true→false. That's the shape, in about the ten lines it should have taken.

What matters more is that you brought tests this time. Last round I wrote *"No new tests were added, which is why I keep having to rebuild this harness each round."* This round I threw three mutants at the new logic and your suite killed all three:

```
D  drop the restore branch entirely        KILLED
E  snapshot an empty object instead        KILLED
F  auto-expand opens no groups             KILLED
```

That's the difference between code that happens to work and code that stays working. `restores collapsed groups after automatic expansion ends` and `toggles the persisted group state repeatedly while filters are active` are both doing real work.

### The coupling — you resolved it the correct way

I'd flagged that narrowing `shouldAutoExpandSidebarGroups` to search-only would silently break mobile, because `ensureSelection` is a second consumer that drives `mobileShowDetail = false`.

You kept the computed broad and fixed the expansion behaviour in the watcher instead. That's the right call — it leaves the mobile contract intact rather than trading one bug for another:

```
B  remove the mobile detail dismissal      KILLED
```

`returns mobile users to the list when a filter becomes active` covers it.

I also checked the logged-out path for a desync, since `activeFilterCount` excludes `currentStatus` when `this.user` is falsy. It's consistent: `filteredProducts` applies the same `this.user ? this.currentStatus : ''` guard, and the status `<fieldset>` is `v-if="user"`, so there's no state where the filter changes the list without changing the count.

### Sidebar counts under filter

`groupedProducts` builds from `filteredProducts`, so `{{ group.items.length }}` is a visible-row count rather than a catalogue count. Defended:

```
A  build the groups from prodList instead  KILLED
```

`sidebarGroupUserToggles` is genuinely load-bearing too — read at line 875 in `syncSidebarGroups`, not write-only state.

### One coverage gap, and it isn't yours

The only mutant that survived:

```
C  delete `this.selectedProductId = this.filteredProducts[0]._id`   SURVIVED  22/22 green
```

That single line is `ensureSelection`'s documented primary job — *"defaults to the first item in the new list"*. Without it, filtering out the selected product leaves `selectedProduct` returning `null`, and since the panel is `v-if="selectedProduct"` the user gets an **empty detail pane with just a back button**. Nothing in the suite notices.

I checked the diff before raising it: that line is **not in your changeset**, so this is a pre-existing gap, not something you introduced. Mentioning it because you now own the best test harness for this page and it's a two-line addition — set a filter that excludes the selection, assert `selectedProductId` moved.

### Gates

```
PR:       322 passed / 3 failed
develop:  318 passed / 3 failed   (same clone, same 3: disable-email-obfuscation, StingChat)
```

Identical pre-existing failures, zero regressions, +4 net tests. Branch is 3 commits behind `develop` but merges cleanly.

### Still not verified, and still worth your eyes

The real-browser question from last round stands: whether the 280 ms animated collapse combined with the immediate `scrollTop` correction produces a visible jump. jsdom has no layout engine, so no test here can answer it.

And the pre-existing crash is still live at this head — five bare `reject` references inside `.catch(e => reject(e))` with no enclosing Promise executor (lines 603, 604, 618, 619, 655/656). Any of those firing throws `ReferenceError: reject is not defined`. It's on `develop`, not yours, but it deserves a card.

Merging. Good turnaround on the tests — that was the thing I actually wanted.

PR #397, opened 2026-08-11
Fix cookie consent banner overlap with proposal modal

reemnii:
Fixes the overlap between the cookie consent banner and DHF proposal popup by dynamically positioning the popup above the banner. Also ensures the layout remains consistent across mobile, light, and dark modes.


mcfarhat (approved):
✅ Approved — clean, well-engineered fix for the cookie-banner / DHF-popup overlap.

What I verified:
- **Approach is solid**: measures the banner height (`getBoundingClientRect().height || offsetHeight`), positions the popup above it via a `:bottomOffset` prop (`bannerHeight + 32/20` for compact/regular). Reacts to changes with a **ResizeObserver** (banner) + a **window resize listener** (viewport).
- **No leaks**: `beforeDestroy` tears down both the observer *and* the resize listener, and resets body state. Guards for missing element / no `ResizeObserver`. SSR-safe (DOM access only in mounted/methods).
- **Banner id matches** (`myPanel1`) between the template, the JS, and the dark-mode CSS.
- **No regressions**: full suite 306 pass / 3 fail — byte-identical to develop's baseline (the 3 are the pre-existing StingChat + email-obfuscation fails).
- **Visually confirmed**: DHF popup now stacks cleanly above the cookie banner with a gap on both **desktop and mobile** — overlap gone.

Nice work. (Please give dark mode a quick eyeball too, since the PR mentions it — the positioning is theme-independent so it should be fine.)

PR #398, opened 2026-08-12
blog screen errors out for non-logged users

ShahdQaddoura:
Fixed the error as asked, when the user clicks on Publish, show a login screen which will allow the user to log in while preserving all the changes they made on the screen.


mcfarhat (changes requested):
Reviewed at `35a18b31`. **The bug you were asked to fix is genuinely fixed**, and you also fixed a leak nobody asked about. Changes requested for the second claim in your description, not for the crash fix.

## The crash fix is correct, and at the right layer

I confirmed the original failure on `develop` by driving `save()` with no session:

```
DEVELOP save() error  -> TypeError: Cannot read properties of null (reading 'account')
DEVELOP loading flag  -> left true
```

`this.editPost.author = this.user.account.name` null-derefs, and because `this.loading = true` is set before the throw and never reset, the Publish button spins **forever** with no error. That's exactly the "errors out" report. Your guard short-circuits before any of it — `save()` no longer throws, `requestLogin()` fires once, `loading` stays `false`, zero store dispatches.

Importantly you fixed the **null-deref**, not the symptom. A `v-if` on the Publish button would have hidden the bug; you found the actual cause.

## Blocking: the draft is not preserved — it's overwritten

Your description says the login "preserv[es] all the changes they made on the screen." I reproduced the opposite.

Your design depends on `loginRequestedByPublish` suppressing `runDraftLoader()` after login. It can't, because of Vue 2's scheduler. `$emit('login-successful')` is **synchronous**; the `user` watcher is **queued to nextTick**. So this always happens in this order:

```
:293 handleLoginSuccessful()  ->  :295 loginRequestedByPublish = false     (sync, from the emit)
:240 watcher user(newUser)    ->  :242 if (this.loginRequestedByPublish)   (nextTick — flag already false)
:247 runDraftLoader()                                                       (runs, clobbers)
```

Verified empirically with a real Vue 2 + Vuex reproduction of both components:

```
1. commit
2. handleLoginSuccessful (clears flag)
3. after-emit
4. watcher:user (flag=false)   <- guard already gone
5. runDraftLoader RAN -> DRAFT CLOBBERED
```

And `runDraftLoader()` (`:964-977`) assigns `title`, `body`, `tags` and `benef_list` unconditionally from `localStorage[<username>_draft_blog]`:

```
title after login -> OLD stale draft title
body  after login -> OLD stale draft body
```

So: an anonymous visitor types a post, clicks Publish, logs in — and their typed content is **replaced by an old abandoned draft**. That's reachable in normal use, not a corner case: `runDraftSaver` writes that key every 20s while logged in, and `$clearDraft` only runs on a *successful* publish (`:615`), so any previously abandoned draft on that account triggers it.

It appears to work only when no stored draft exists — `$loadDraft` returns `null`, `JSON.parse(null)` is falsy — which is almost certainly the path you tested.

**Suggested fix, and it's simpler than what's there now:** drop the flag entirely and make `runDraftLoader()` non-destructive — only apply a stored field when the current in-memory field is empty. That removes the need to sequence two async paths against each other, which is what made this fragile in the first place.

## Blocking: cancel the prompt, then log in elsewhere, and autosave dies

`handleLoginClosed()` (`:290-292`) sets `showLoginModal = false` but never resets `loginRequestedByPublish`. Reproduced:

```
flag after cancel     -> true
startDraftSaver calls -> 0
runDraftLoader calls  -> 0
```

Path: type → Publish → dismiss the modal → log in via the navbar instead. NavbarBrand has its **own** `LoginModal` instance, so it emits to NavbarBrand, not to this page — `handleLoginSuccessful()` never runs, the flag stays `true`, the watcher early-returns, and `startDraftSaver()` is never called. **The draft is never persisted for the rest of the session.** One line in `handleLoginClosed()` fixes it, and it disappears entirely under the non-destructive-loader approach above.

## The shared component change is safe — I checked all of it

You added `$emit('login-successful')` to the hiveauth (`:147`) and keychain (`:150`) handlers; the posting-key path already had it. So it's a consistency fix, not new API.

Of the **14** consumers of `LoginModal`, only **three** bind `@login-successful`, so the blast radius is two pre-existing handlers:

- `NavbarBrand.vue:123` — now also fires for keychain/hiveauth, causing one redundant `refreshUser` dispatch (LoginModal already dispatches it). Benign.
- `wallet.vue:1860` — keychain/hiveauth users now get `refreshAllWalletData` after login where they previously didn't. That's arguably a latent bug being fixed, and the method is idempotent.

The other 11 bind only `@close`, so they're provably unaffected. No regression found.

## Non-blocking

- **Match the house jQuery guard.** Every other consumer wraps the show call: `this.$nextTick(() => { if ($ && typeof $.fn.modal === 'function') { ... } })`. Yours (`:284-288`) has the `$nextTick` but omits the guard. jQuery comes from a CDN (`nuxt.config.js:251`), so if it's slow or blocked this throws and Publish silently does nothing — straight back to a dead button, which is the bug you're fixing.
- While you're in `runDraftLoader`: `JSON.parse(data)` is unguarded and `$loadDraft` returns `''` if the localStorage read throws — and `JSON.parse('')` throws. Not yours and not currently reachable, but cheap hardening since you're already there.

## Auth: no weakening, stated plainly

This only changes *when* the login prompt appears. Your guard is strictly **more** restrictive than develop, which crashed instead of proceeding. Real authorisation is untouched and lives downstream at `:854+` — broadcast still requires keychain signing, hiveauth, or a server-side token. Nothing here lets an unauthenticated user post.

## Gates — identical to develop

`npm ci` exit 0, `npm run lint` exit 0, `nuxt build` exit 0. Jest: 3 failed / 306 passed, and **identical on `develop`** — the same two known suites. Zero new failures.

## Why it shows BLOCKED — nothing for you to do

Not conflicts, not behind develop, not a failing check, and not a stale review. `mergeable: MERGEABLE`, 0 commits behind, the required `npm ci` check passes, and there are **zero reviews** — protection just requires one approval. Note `require_last_push_approval: true`, so the approval has to come *after* your next push.

## Overlap with #377 — no conflict, one thing to watch

File sets are disjoint, so no merge conflict. But they meet at one point: `wallet.vue:1860` is one of the two `@login-successful` consumers, and #377 rewrites `wallet.vue` substantially. Whichever merges second, we should confirm the binding survived the rewrite and that `refreshAllWalletData` behaves on a keychain login. Not blocking this PR.

## What you did well

**You fixed a real leak you weren't asked about.** `develop` called `setInterval(this.runDraftSaver, 20000)` in *both* `mounted` and the `user` watcher with no cleanup — measured at **3 stacked intervals** after 3 user changes, and none cleared on unmount. Your `startDraftSaver()` idempotence guard plus the new `beforeDestroy()` `clearInterval` closes both. That's a genuine bug found in passing.

You also followed house conventions closely — the `$('#loginModal').modal('show')` inside `$nextTick` after a `v-if` mount matches the other 11 consumers, and putting `LoginModal` alongside `NavbarBrand` mirrors 9 existing pages, so you didn't invent a pattern. And you made the shared component consistent across all three login methods rather than only the one you needed.

Small focused diff, lint and build clean first time, no new test failures. Fix the draft-preservation path and this is done.

mcfarhat (changes requested):
## Re-reviewed at `5e027d6b` — everything I asked for is fixed. One item left, in the new video code.

All four previous items verified by execution, with negative controls so I know the checks could have failed. I drove the real component options on a real Vue 2 + Vuex instance with a real jsdom `<select>` — 30 assertions, 30 green including 5 negative controls.

### The draft-clobber fix — closed, and you took the better route

You deleted `loginRequestedByPublish` entirely (`grep -r` across the repo: zero hits) and made `runDraftLoader()` non-destructive, rather than trying to sequence against Vue's scheduler. That removes the sync-`$emit`-vs-`nextTick`-watcher race instead of fighting it, which is why blocking items 2 and 3 both closed with one change.

Driven through the real scheduler — type content, `save()`, commit user, synchronous `handleLoginSuccessful()`, two `nextTick`s:

```
runDraftLoader spy      -> CALLED            (proves the watcher fired — not a vacuous pass)
title                   -> "MY TYPED TITLE"  (stale draft was "OLD DRAFT TITLE")
description             -> "MY TYPED DESCRIPTION"
$refs.editor.content    -> "MY TYPED BODY"
benef_list              -> []                (stale draft had bob@500)
```

Then I swapped develop's destructive loader back in and re-ran the identical sequence: `title === "OLD DRAFT TITLE"`, `body === "OLD DRAFT BODY"`. Red. The check is live.

Also closed: the null-deref in `save()` (logged-out now returns cleanly, `showLoginModal=true`, `loading` stays `false`); the autosave-death path (with the flag gone there's nothing to leak — `draftSaveInterval` survives a cancelled prompt); the jQuery guard; and the unguarded `JSON.parse`, where you went further than asked and array-guarded `tags`/`beneficiaries` too.

The interval-leak fix from the earlier round still holds: 3 user changes → 1 timer, `$destroy()` → 0. Two negative controls both went red — strip your `beforeDestroy` and the timer outlives unmount; restore develop's unguarded `setInterval` and you get **3 stacked timers**, matching the original measurement.

And `description` round-trips correctly with #399 — full save→load of title/description/body/tags/beneficiaries/community is identity, and a fresh logged-in load with an empty composer still restores every field. No regression for logged-in users.

`LoginModal` blast radius re-checked at this head: the diff is 2 added `$emit('login-successful')` calls and nothing else, `LoginModal.props` is empty so there's no consumer contract to break, and a failed keychain login correctly does **not** emit.

---

## Blocking — the video guard is correct but unreachable

`fix video publishing for logged-out users` adds the right guard in the right shape, mirroring the blog one. The problem is that a logged-out user can never reach the code it protects, so the commit doesn't fix the bug it names.

The Publish button is `:disabled="!selVidReadyForPublish()"` (`pages/_username/videos/new.vue:185`), and:

```js
selVidReadyForPublish() {
  return (this.selVid && this.selVid.status && this.selVid.status == 'publish_manual');
}
```

`selVid` defaults to `{}`, and a logged-out user can never populate it, because the whole upload/video-list card is `v-if="xcstkn != ''"`. `xcstkn` is only ever set by `connectSession3S()`, which null-derefs around line 1588:

```js
let res = await client.get(process.env.threeSpeakApiSession.replace('_USERNAME_', this.user.account.name), ...)
```

and its own `catch (exc)` at ~1728 swallows the `TypeError`. So `save()` is dead code for exactly the case it was written for.

Executed in jsdom, all assertions passed:

1. Load `/@anyone/videos/new` logged out → `mounted()` → `await this.connectSession3S()`
2. Throws `TypeError: Cannot read properties of null (reading 'account')` → swallowed by its own catch
3. Observed: `xcstkn === ''`, `loadingxcstkn === false`, `userVidList === []`, **`showLoginModal === false`**, no error escapes
4. Template renders only the fallback `<button @click="connectSession3S">Load Videos</button>`
5. Click "Load Videos" → same swallowed `TypeError`. **Nothing happens.** No login prompt, no message, no spinner.

That silent dead-end *is* the logged-out video bug, and it's untouched.

**The fix belongs in `connectSession3S()`** (and/or the "Load Videos" click handler): guard `!this.user?.account?.name` and call `requestLogin()`, the same shape as `save()`. Keep the `save()` guard — it's correct defence in depth and consistent with the blog page. It just isn't the fix.

What the video commit *does* get right, and I verified it: `user(newUser) { if (!newUser) return; ... }` correctly stops `connectSession3S()` re-entering with a null user on logout, and after a successful login the watcher does fire `fetchCommunities()` + `connectSession3S()`, so the page self-heals once logged in.

---

## Non-blocking

- **Partial-fill mixed state.** Type a body, leave the title blank, log in → the stale draft's title/tags/beneficiaries arrive alongside your typed body. Verified: `editor.content === "MY TYPED BODY"` but `title === "OLD DRAFT TITLE"`. That's exactly the "only fill empty fields" semantics I asked for, so it's expected — but worth one comment in `runDraftLoader` so the next reader doesn't file it as a bug.
- **Publish isn't retried after login** — `save` is not re-invoked, so the user clicks Publish a second time. Fine, just unpolished.
- **`handleLoginSuccessful()` duplicates the watcher.** Measured `fetchCommunities` firing exactly **2×** per login, once from the handler and once from the watcher on `nextTick`. `startDraftSaver` also twice, but it's idempotent. The handler body could shrink to `this.showLoginModal = false`.
- **`targetCommunity` restore race** (pre-existing, not yours): if `runDraftLoader` runs before `fetchCommunities()` resolves, the `<select>` drops the unknown value. I specifically checked whether your new `=== '_blog_'` guard makes that permanent — **it doesn't**; the select resets to its first option and a later `runDraftLoader` recovers it. Behaviour matches develop.

## One thing that isn't yours

`develop` is currently **red**: `test/unit/seo/internal-navigation.spec.js` gained 2 failures from #401, which changed `NavbarBrand` to locale-aware links while the spec still asserts the literal `href="/blog/new"`. I ran it on a clean `origin/develop` worktree and reproduced it there. So the baseline is now 5 failures, not 3 — and **this PR adds zero**:

```
PR  5e027d6b   5 failed / 304 passed / 309
develop d04ec3c  5 failed / 304 passed / 309   (identical)
```

Mentioning it so a red run doesn't send you hunting. I'm fixing that separately.

Gates on this head: `npm ci` exit 0, `npm run lint` exit 0, `npx nuxt build` exit 0.

You also kept the branch current through three same-day merges into develop, so `description` round-trips and there was no merge risk when I reviewed. Note it's gone `BEHIND` again since the #377 merge — worth syncing before the next round.

---

Fix the `connectSession3S()` guard and this is done. Taking the architectural fix on the draft clobber rather than patching around the scheduler was the right call, and it's why this round was short.

mcfarhat (approved):
## Approved at `04aa2386`

The blocker is closed, you rebased cleanly, and the gates are identical to develop. Merging.

First, a correction to my own review — **my diagnosis of the mount-time path was wrong.**

I told you `mounted()` reaches `connectSession3S()`, null-derefs on `this.user.account.name`, and its own `catch` swallows the `TypeError`. Driven against the real store module, it never gets that far:

```
MOUNT/logged-out  connectSession3S calls = 0
                  fetchCommunities calls  = 0
                  showLoginModal = false, xcstkn = "", unhandled rejections = 0
```

The real cause is in `vue-steemconnect/src/index.js`: the `login` action returns a promise whose executor only calls `resolve()` if `state.user` is already set or an `access_token` exists. For a logged-out visitor neither holds, so **the promise never settles** and `mounted()` stalls at `await this.$store.dispatch('steemconnect/login')` — before your guard is ever reached. Confirmed by negative control: seed an `access_token` and the same dispatch resolves and `connectSession3S` runs.

That's byte-identical on develop and untouched by your PR, so it's neither yours nor a regression. Net effect is that no modal auto-pops on page load, which is arguably better behaviour, and the dead-end is gone because the click path works. Not blocking — but you should know the mechanism I described wasn't the real one.

### What's verified

I drove the real SFC in jsdom with the real Vuex store and real handlers — 27 assertions, 6 of them negative controls that I confirmed go red.

**Load Videos while logged out — closed:**
```
CLICK  showLoginModal = true,  LoginModal in DOM = 1,  unhandled errors = 0
NEG CONTROL (guard stripped)  showLoginModal = false, xcstkn = ""   <- red
```

**Recovery after login — closed, and it fully recovers.** I ran the genuine first-login path — no cached 3Speak token, posting-key login through server-side `memoDecode` — not just the localStorage shortcut:

```
xcstkn        = "3S_TOKEN"   (leading '#' stripped)
userVidList   = [{_id:"v1", status:"publish_manual", ...}]
"Load Videos" still shown = false   (upload/list card now renders)
```

That was the thing I was most concerned about: a guard that prompts and then leaves the page dead is only half a fix. It genuinely self-heals.

**Defence in depth intact.** `save()` still guards, and the `user` watcher still has `if (!newUser) return` — on logout, `connectSession3S` calls = 0 and `requestLogin` calls = 0. Negative control: strip that guard and logging out pops the login modal in the user's face. **No regression for logged-in users** — the card renders, and `save()` proceeds past the guard into the real body, so it isn't over-restrictive.

**Blog side still holds at this head:** `loginRequestedByPublish` is 0 hits repo-wide; typed content survives login against a stale draft; the interval leak is still closed (3 user changes → 1 timer, `$destroy()` → 0, with develop's version giving 3 stacked and 3 surviving). `description` from #399 round-trips on both the page you touched and the blog composer.

### Gates

```
PR head   3 failed / 306 passed / 309     npm ci exit 0, lint exit 0, nuxt build exit 0
develop   3 failed / 306 passed / 309     (same node_modules)
```

Identical, and the baseline is 3 rather than 5 now that the `internal-navigation` pair is fixed. Zero added.

You also rebased onto current develop yourself — `merge-base` is the develop tip, trial merge says "Already up to date", zero conflicts. Nothing to untangle.

### Two follow-ups, neither yours to fix here

**`loadingxcstkn` never resets in the cached-token branch** (pre-existing, identical on develop). Measured `true` after a successful cached load, so any later `connectSession3S()` hits `if (this.loadingxcstkn) return` and silently no-ops. Low impact today because the card is already rendered by then — but it's the same class of silent dead-end you just fixed, so it's worth its own card.

**Duplicate DOM id:** the video page now renders a second `LoginModal`, whose root is `id="loginModal"`, and `NavbarBrand` renders one too. Only reachable with the navbar modal already open, and the blog page already ships this pattern — a nit, noted so it isn't a surprise later.

### What I couldn't check

Bootstrap backdrop teardown in a real browser when `v-if` removes the modal mid-hide — jsdom has no Bootstrap transitions. Reading it, the sequencing is right (`modal('hide')` before the `login-successful` emit, `close` emitted from `hidden.bs.modal`), and it matches `NavbarBrand`'s pattern, so it's the house convention. Worth an eye on a real device when you check the card.

Putting the guard in `connectSession3S()` rather than hiding the button was the right layer, and keeping `save()` as defence in depth was exactly what I asked for. Two commits, four lines of real change.

PR #403, opened 2026-08-15
made follow-up changes requested

ShahdQaddoura:


mcfarhat (changes requested):
## Changes requested at `e0285691` — I think we're talking about two different lists

The PR title says "made follow-up changes requested", and you've done a real round of work — but it's the **11 August presentation round** (action-box layout, close affordances, top-bar cleanup), not the four items I carded from the #377 wallet review. Both got called "follow-ups", so this reads as a genuine mix-up rather than anything skipped, and the ambiguity is mine.

**The four from the wallet review, none of which this PR touches:**

1. **VESTS → HP in the HIVE history.** `normalizeHiveTransaction` still does `amounts.join(' + ')` on raw op strings. Driven with a real `curation_reward`: renders `"203.521707 VESTS"`. Live mainnet ratio is `214354619.506 / 346049743609.085690 = 0.00061943…`, so that's **0.126 HP** — the displayed figure is **1,614×** the HP number beside it on the same page.
2. **`balance-display-key` vs `balanceDisplayNameKey`.** Still 3 call sites wrong, 0 right. Proven by asserting the declared prop rather than the stray attribute: `vm.balanceDisplayNameKey === "Balance"` (the default), while `vm.$attrs === {"balance-display-key":"HP_Bal"}` and leaks onto the DOM. Mutation confirms — pass the correct name and the prop reads `HP_Bal` with `$attrs` empty.
3. **`TopHoldersXModal.vue`** still has no `holders-table` class and no dark-mode rule; its sibling has both.
4. **`account_est_val`** still renders twice.

Worth saying plainly: **the work you did do is fine**, and I'd rather have it than not. This is about which list, not about quality.

---

## New defect — a stray Close button on a loaded wallet

`wallet.vue:729` guards on `fundActivityMode !== CLOSED_MODE`. But `data()` line 2127 sets `fundActivityMode: this.CLOSED_MODE`, and `CLOSED_MODE: 0` is declared at line 2085 **inside that same `data()` return literal** — so `this.CLOSED_MODE` is `undefined` at that point and `fundActivityMode` initialises to `undefined`, not `0`. `undefined !== 0` is true, so the guard passes with nothing open.

Executed on a loaded owner wallet:

```
fundActivityMode = undefined | CLOSED_MODE = 0 | guard = true
wallet-action-dismiss = 1   open core panels = 0
```

A Close button floating over the panel area with nothing to close. Mutation: force `fundActivityMode: 0` → **0** occurrences. Control: `fundActivityMode: 1` → present, correctly.

The `undefined` init is pre-existing — but it was **inert until this PR added a guard that compares against it**. And here's the thing: your own `isCoreAfitActionOpen` / `isAfitActionOpen` computeds *don't* have this flaw, because `.includes(undefined)` is simply false. You got two of three guards right by using the better pattern, then hand-wrote the third as a raw `!==`. Using the same declarative shape for all three fixes it.

## A product decision I need from you, not a bug

You removed four toolbar buttons — `showClaimableRewards`, `fetchDelegations`, `scrollToTransactionHistory`, `showBSCAddress` — as duplicates of the #377 quick-action pills. Driving real clicks with spies, **all four still fire for the owner**, so no loss there.

But the pills nav is `v-if="!nonAuthUser"`, whereas on develop those four toolbar spans carried **no auth guard at all**. So for a logged-out visitor, or anyone viewing someone else's wallet, all four become unreachable — including `scrollToTransactionHistory` and `fetchDelegations`, which are read-only and arguably public-interest. Deleting the duplicates is correct de-duplication; losing the logged-out path may not be intended.

---

## Two more before merge

**i18n.** `No_data_found_for_top_holders` was added to **en_US only** — missing from the other 13 locales. The convention (`off_chain` in #377) is all 14 with real translations.

**`nuxt.config.js` scope creep, and this one is actively harmful in dev.** 18 lines unrelated to any follow-up — the same split-it-out note I gave you on #377. Two concrete problems:

- The existing dev proxy entries are **relative** (`'/steem-api/rpc'`); yours hardcode **absolute `http://localhost:3000/...`**. If the dev server isn't on 3000 — `PORT` set, or the port taken — every Actifit-API and Hive-Engine call in dev breaks.
- `actiAppUrl` in dev now defaults to the proxy → `https://api.actifit.io`, where it was `localhost:3120`. So a developer with no env file is now talking to **production** by default.

Production builds are unaffected (`NODE_ENV === 'production'`), but that's a nasty surprise for whoever picks this up next.

Also: `<h5>{{ $t('wallet.initiate_afit_to_he') }}</h5>` was removed with no explanation, and the PR description is empty with no screenshots on an entirely visual change.

---

## What held up

**The money is right, and I checked it numerically rather than trusting the render.** HP modal against live `api.syncad.com` plus live global props, recomputed independently as `value/1e6 × ratio` — top 10 exact to a relative error under 1e-12: blocktrades **16,389,042.554 HP**, freedom 14,308,931.687, alpha 11,831,554.420. Mutation: drop the `/1e6` and everything shifts by over 90%. Wallet page against `condenser_api.get_accounts` for `actifit`: `83.416 HIVE` and `9.218 HBD`, both byte-exact on chain. No conversion path was touched and nothing drifted.

**Logged-out safety passes**, and the check is honest — my first version searched the rendered HTML for handler names, which is vacuous in Vue 2. The sanity control caught it: the same procedure on the owner view fires 6 handlers, so the check could have failed. On the logged-out view it fires **zero**.

All #377 work intact: `holders_result` envelope, `apiCoinType` returning `VESTS`, the `v-if="!nonAuthUser"` nav guard, the hero `-1` guard, and token sorting all survive mutation.

Gates: `npm ci` 0, lint 0, `nuxt build` 0, jest **3 failed / 306 passed** — exactly baseline, zero added. And you merged develop in yourself, so you're 0 behind with no conflicts.

**You also fixed a real bug nobody carded:** `No_data_found_for_top_holders` was referenced by `GenericTopHoldersModal` but absent from `en_US`, so the empty state was rendering the raw key. Good catch.

## Two things for cards, pre-existing and not yours

`wallet.vue:3767` — `renderBalance` returns **JSX** when `displayUserData.balance` is missing, and line 186 interpolates it as `{{ }}`. Interpolating a VNode calls `JSON.stringify` on it → `TypeError: Converting circular structure to JSON` → **the entire wallet page render aborts**. I hit this live. That same JSX is also why `wallet.vue` can't be imported by the repo's jest config at all, which is why no wallet spec exists.

## Not verified

Anything visual — jsdom applies no CSS. The actual appearance of the Close buttons, toolbar wrapping, whether `.wallet-action-dismiss` overlaps panel content, dark mode and mobile all need a real browser. Since this PR is entirely presentation, that pass is required before it merges.

---

Sorry about the ambiguity on "follow-ups" — that's on me for using the same word for two lists. Fix the stray Close button, add the 13 locales, split out the `nuxt.config.js` change, and tell me what you want to do about the four toolbar buttons for logged-out users. The four wallet-review items can be a separate PR or this one, whichever you prefer.

ShahdQaddoura:
Hey, what I worked on was the last 5 comments you mentioned yesterday on Trello. Since you closed the pr on the last branch so I just thought it was better to make a new branch and commit the 5 changes, but sure, I will work on the ones you just mentioned! Sorry for the confusion; I forgot to leave a description to explain what I committed.

ShahdQaddoura:
Okay, so I have addressed the four wallet-review items, fixed the stray Close state, translated across all 14 locales, and removed the unrelated nuxt.config.js proxy change. For the four duplicated toolbar actions, I thought of keeping the lower controls available for logged-out visitors or visitors viewing another wallet only. At the same time, owners use the top pills, which preserves the public behavior from develop without duplicating controls for owners. Does this match the intended product behaviour?

mcfarhat (changes requested):
## Re-reviewed at `9494d8c9`

The dismissal work is done properly. One blocking issue in the new VESTS→HP conversion.

### Blocking — VESTS→HP can silently overstate by ~1,600x

`formatHiveHistoryAmount` (line 2974) multiplies by `hivePerVestsRatio`, whose fallback is `1`:

```js
return 1; // Default to 1 if properties aren't loaded or chain is not HIVE, or division by zero
```

That fallback was written for the *other* consumer of this computed — `:divisor="hivePerVestsRatio"` at line 1945, where `1` is a safe identity for a division. In a multiplication it is not safe: it passes the raw VESTS figure straight through and relabels it `HP`.

Against mainnet global properties right now:

```
total_vesting_fund_hive : 214365804.479 HIVE
total_vesting_shares    : 346058997843.434021 VESTS
ratio                   : 0.00061945 HP per VEST

a "1000.000000 VESTS" row renders as
  ratio loaded   : 0.619 HP
  ratio = 1      : 1000.0 HP      <-- overstated 1,614x
```

**It is reachable**, because nothing orders the two loads:

```
data()   properties: ''                              <- falsy, so ratio = 1
mounted()          this.loadGlobalProperties()       <- not awaited  (line 8640)
fetchUserData()    this.fetchHiveFinancialHistory()  <- not awaited  (line 3919)
```

Between `fetchUserData` starting and line 3919 nothing assigns `this.properties`, so the two are a straight race.

**And a lost race is permanent, not a flicker.** Line 2914 assigns the mapped result into `data`:

```js
this.hiveTransactions = (Array.isArray(history) ? history : []).map(entry => { ... })
```

`formatHiveHistoryAmount` runs once during that map and the converted *string* is stored. It is not a computed, so when `properties` resolves a moment later there is no reactive dependency to re-evaluate — the wrong figures stay on screen for the life of the page view.

The reason this will look fine when you test it: `getAccountHistoryAsync(account, -1, 500)` is a far heavier call than a single `get_dynamic_global_properties`, so properties normally wins and the numbers are right. It fails on a slow or congested connection — intermittently, on a wallet page, in the user's favour-looking direction. That is the worst shape a display bug can have.

Two ways to close it, either is fine:

1. `await this.loadGlobalProperties()` before the history maps, or
2. keep the raw amount in `hiveTransactions` and convert in a computed, so it self-corrects when the ratio arrives.

Either way **the fallback should refuse to convert** rather than mislabel — return the original `amount` string (still marked `VESTS`) when `properties` is unloaded. A visibly unconverted row is a much smaller failure than a confidently wrong one.

### Minor — decimal formatting

```js
return Number(hp.toFixed(3)) + ' HP';
```

`Number()` strips trailing zeros, so this renders `1.5 HP` where the rest of the wallet shows three decimals. Drop the `Number()` wrapper.

### Verified working — the dismissal work

I checked this properly rather than eyeballing it:

- All four `wallet-action-close` buttons (599, 729, 1333, 1385) sit behind guards, and the guards are deliberately disjoint — line 729 carries `&& afitActivityMode !== showBSCDetails` specifically so it can't double up with the BSC panel's own button at 1333. That's the right instinct.
- `closeAllWalletActions` resets to `CLOSED_MODE: 0`, and I checked every mode constant for a zero collision: fund modes are 1-6, 10, 11, 13, 14 and AFIT modes 1-9, 12. Nothing else is `0`, so closing genuinely closes.
- `isCoreFundActionOpen` covers all ten declared fund modes — no mode can open a panel that has no dismiss control.
- The `Close` key is present in **all 14 locale files**, correctly translated. Nice — the usual failure here is `en_US` only and a raw key showing in the other thirteen.

Dropping the `console.error("Element not found.")` is right too; that fired on every page without a details area.

### Housekeeping

The branch is **9 commits behind `develop`** (3 ahead, `mergeStateStatus: BEHIND`). Please merge `develop` in before this goes anywhere.

### Method note

The arithmetic above is live mainnet data and the call-graph claims are read directly off `9494d8c9`; I did not build and drive the Nuxt app for this round, so the race is established by ordering rather than observed failing. The ordering isn't ambiguous, but calling it as I found it.

mcfarhat (approved):
## Re-reviewed at `84651bfc` — fixed, verified by execution. Approving and merging.

`84651bfc` closes the blocking issue properly, and it closes it in all three places rather than just the one I pointed at.

### The fix

```js
if (!this.hasHivePerVestsRatio()) {
  await this.loadGlobalProperties();      // 1. ordering: no longer a race
}
...
if (!this.hasHivePerVestsRatio()) return amount;   // 2. refuses to convert
const hp = parseFloat(value) * this.hivePerVestsRatio;
return hp.toFixed(3) + ' HP';                       // 3. Number() wrapper gone
```

Awaiting the properties removes the race, and the guard means that even if `loadGlobalProperties()` fails outright — offline, RPC down — the row degrades to an honest `1000.000000 VESTS` rather than a confident `1000.0 HP`. That's the belt-and-braces version of what I asked for: correct ordering *and* a fallback that can't lie.

### Verified by running your code, not reading it

`wallet.vue` can't be imported into Jest at all — `pages/_username/wallet.vue:3802` contains real JSX (`return <i class="fas fa-spin fa-spinner text-white"></i>;`) and the Babel config has no JSX plugin, so the whole file fails to parse. Pre-existing and nothing to do with you, but it meant I couldn't write a normal spec. Instead I extracted `formatHiveHistoryAmount`, `hasHivePerVestsRatio` and `hivePerVestsRatio` verbatim from the file at this commit and drove them directly:

```
PASS  properties loaded: converts        -> "0.619 HP"
PASS  UNLOADED: no HP mislabel           -> "1000.000000 VESTS"
PASS  zero vesting shares: no convert    -> "1000.000000 VESTS"
PASS  3 decimals kept                    -> "1.500 HP"
PASS  non-VESTS untouched                -> "5.000 HIVE"
PASS  non-HIVE chain: no convert         -> "1000.000000 VESTS"
PASS  malformed shares: no convert       -> "1000.000000 VESTS"
```

The `0.619 HP` is against live mainnet globals, so that's the real number a 1000 VESTS row now renders.

And I mutation-tested my own harness so it isn't just agreeing with you — deleting the guard line reproduces the original defect exactly:

```
guard removed  -> "1000.000 HP"      <- the 1,614x mislabel, back
guard restored -> "1000.000000 VESTS"
```

Two cases I added beyond what I originally reported, both of which you handle: a non-HIVE chain (`cur_bchain !== 'HIVE'`), and a malformed `total_vesting_shares` that parses to `NaN`. `NaN > 0` is false, so it falls through to the no-convert path rather than producing `NaN HP`.

### Gates

```
PR:       325 passed / 3 failed
develop:  325 passed / 3 failed   (same clone, same 2 suites)
```

No regressions. The branch is also no longer behind — `c7e7241e` merged develop in, so `BEHIND` is cleared and only the review gate remains.

### One small note, not worth changing now

`hasHivePerVestsRatio()` duplicates the condition inside the `hivePerVestsRatio` computed. They agree today; if someone edits one later they can silently diverge, and the divergence would reintroduce exactly this bug. Deriving the guard from the computed (e.g. having the computed return `null` when unavailable and checking for that) would make them impossible to disagree. Not blocking — the current version is correct and I verified both branches of it.

Good, thorough fix. You went past the specific line I flagged and covered the failure modes around it.

PR #404, opened 2026-08-15
Fix/align post comment actions

reemnii:
Follow-up cleanup restores comment-image containment, removes dead scoped footer CSS, restores semantic keyboard-accessible reply controls, and adds mounted Comments/CardActions integration coverage.

### DIFF
--- a/components/CardActions.vue
+++ b/components/CardActions.vue
@@ -13,9 +13,9 @@
       <a href="#" class="post-detail-action" :class="{ 'post-detail-action--active': hasVoted }" @click.prevent="$emit('vote-prompt')" :data-toggle="voteModalTarget ? 'modal' : null" :data-target="voteModalTarget" :title="$t('votes')">
         <i class="far fa-thumbs-up"></i> {{ voteCount }}
       </a>
-      <a href="#" class="post-detail-action" v-if="showComments && !commentsDisabled" :class="{ 'post-detail-action--active': commentsActive }" @click.prevent="$emit('open-modal')" :data-toggle="modalTarget ? 'modal' : null" :data-target="modalTarget" :aria-expanded="commentsActive ? 'true' : 'false'" :title="commentsTitle || $t('comments')">
+      <button type="button" class="post-detail-action" v-if="showComments && !commentsDisabled" :class="{ 'post-detail-action--active': commentsActive }" @click="$emit('open-modal')" :data-toggle="modalTarget ? 'modal' : null" :data-target="modalTarget" :aria-expanded="commentsActive ? 'true' : 'false'" :title="commentsTitle || $t('comments')">
         <i class="far fa-comments"></i> {{ commentsCount !== null ? commentsCount : cardData.children }}
-      </a>
+      </button>
       <span v-else-if="showComments" class="post-detail-action post-detail-action--static">
         <i class="far fa-comments"></i> {{ commentsCount !== null ? commentsCount : cardData.children }}
       </span>
@@ -80,6 +80,9 @@ export default {
   display: inline-flex;
   align-items: center;
   gap: 5px;
+  padding: 0;
+  border: 0;
+  background: transparent;
   color: var(--post-footer-muted, #333);
   text-decoration: none;
   transition: color .15s ease;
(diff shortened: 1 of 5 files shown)


mcfarhat (approved):
## Reviewed at `dcb9fa72` — approving. The fix is right; two of the assertions can't prove it.

### The change is correct, and for the right reason

Swapping `<a href="#">` for `<button type="button">` is the actual fix, not a cosmetic one. An anchor fires its handler on **Enter but not Space**; a native button fires on both. So keyboard users could never activate the comments toggle with the key most of them reach for first. Using the native element rather than bolting `@keydown.space` onto the anchor is the correct call — you inherit activation, the implicit `button` role, and form-control focus semantics for free, instead of reimplementing three of them.

Dropping `.prevent` along with it is right too — there's no navigation left to prevent.

I verified the rest by mutation rather than by eye (`npx jest`, 12/12 green at baseline):

```
B  revert <button> back to <a href="#">   KILLED  (3 failures)
C  remove the :aria-expanded binding      KILLED  (2 failures)
```

So the button-vs-anchor guarantee and the `aria-expanded` toggle are both genuinely defended. The `Comments.spec.js` integration test earns most of that: mounting with `stubs: { CardActions: false }` so it drives the **real** child component, then asserting `aria-expanded` flips `'true'` → `'false'` across a click, is exactly the right shape. That test is the strongest thing in this PR.

### `.cmt-body img` — the trap you avoided

`Comments.vue`'s `<style>` block is **unscoped**, which is what makes `.cmt-body img` work at all. The images come from `SafeRemarkable` rendering user markdown, so they never carry a `data-v-` scope attribute; had that block been `scoped`, the rule would have silently matched nothing. Worth knowing you got this right on purpose or by luck, because it's the single most common way this fix fails.

The CSS you deleted from `pages/_username/_permlink/index.vue` is also safe to delete — that file is `<style scoped>`, so `#main-footer .post-detail-action` was never reaching the inner elements of a child component anyway. `CardActions.vue` now supplies the equivalent (`gap: 16px`, `flex: 1 1 auto`, `min-width: 0`), so there's no layout regression. Net dead-code removal.

### Two assertions that can't fail

This is the part to take away, and it's a pattern rather than a nitpick.

**1. The focus indicator.** I set `outline: none` and left the selector in place:

```css
.post-detail-action:focus-visible {
  outline: none;          /* <- indicator destroyed */
  outline-offset: 3px;
}
```

```
A  neuter the focus outline, keep the selector   SURVIVED   12/12 still pass
```

`expect(componentSource).toContain('.post-detail-action:focus-visible')` only proves that *string* is somewhere in the file. The test is named "uses a button with an accessible focus indicator" but the indicator can be completely removed without it noticing.

**2. The image constraint.** Same shape. I gutted the rule to `display: block` and moved the declarations to an unrelated selector elsewhere in the file:

```
D  gut .cmt-body img, put max-width/height on another rule   SURVIVED   8/8 pass
```

The three `toContain` checks are independent substring searches across a 1,300-line file — nothing anchors them to the same rule. It happens to be green today only because each string occurs exactly once.

**I'm not asking you to fix these**, and that's deliberate: jsdom doesn't apply SFC `<style>` blocks at all, so no unit test in this suite can assert a computed outline or a rendered `max-width`. The honest options are to keep them as deletion canaries while *naming* them for what they are, or drop them and cover it in a browser-based check. What I'd push back on is the current naming — a test called "accessible focus indicator" that survives the indicator's removal will be trusted by the next person to touch this file. That trust is the actual cost.

### Follow-up, not blocking

Five siblings in the same component still carry the defect you just fixed — edit, delete, reply, vote and reblog are all `<a href="#" @click.prevent>` (lines 4, 7, 10, 13, 22). Same missing Space activation, same anchor-with-no-href-target. Out of scope for a PR titled "align post comment actions", but they're one mechanical pass away and worth a card.

### Suite state

`Tests: 3 failed, 318 passed`. I checked those 3 against `origin/develop` and they are **identical and pre-existing** (`disable-email-obfuscation.spec.js`, `StingChat.spec.js`, 3 failed / 315 passed there). Nothing here regressed; your 3 new tests are the delta. Branch is 0 behind develop.

Approving and merging — the accessibility fix is correct and shippable, and I'd rather land it than hold it over assertion wording.

